PR #22, opened 2020-12-04
Add push notification reminder

raymondjacobson:
https://trello.com/c/vWEe1wwb/1693-reminder-to-turn-on-notifications

Add a drawer that shows up every 10 sessions to have a user turn on notifications if they've never been turned on or explicitly disabled. Also shows on native mobile sign in flow.

New tools:
- ReactNativePermissions library (our react native push notifications doesn't have a way to handle the notification state "unset but requestable" (see Results.DENIED) in the PR)
- Session counter
- Lifecycle state for signed in

Tests
- [x]  Install app → after sign in, shows drawer to enable notifs
- [x]  If drawer dismissed, quitting and opening the app 10 times shows the drawer
- [x]  If accepted, quitting and opening the app 10 times does not show the drawer
- [x]  If denied, quitting and opening the app 10 times does not show the drawer

### DIFF
--- a/src/components/notification-reminder/NotificationReminder.tsx
+++ b/src/components/notification-reminder/NotificationReminder.tsx
@@ -1,0 +1,59 @@
+import React, { RefObject, useCallback } from 'react'
+import { checkNotifications, RESULTS } from 'react-native-permissions'
+
+import { MessagePostingWebView } from '../../types/MessagePostingWebView'
+import { postMessage } from '../../utils/postMessage'
+import useSessionCount from '../../utils/useSessionCount'
+import { MessageType } from '../../message'
+
+const REMINDER_EVERY_N_SESSIONS = 10
+
+type OwnProps = {
+  webRef: RefObject<MessagePostingWebView>
+  isSignedIn: boolean
+}
+
+const NotificationReminder = ({ webRef, isSignedIn }: OwnProps) => {
+    
+  const remindUserToTurnOnNotifications = useCallback((count: number) => {
+    if (webRef.current && isSignedIn) {
+      checkNotifications().then(({status}) => {
+        switch (status) {
+          case RESULTS.UNAVAILABLE:
+            // Notifications are not available (on this device / in this context).
+            // Do nothing.
+            return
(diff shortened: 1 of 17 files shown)


src/utils/useSessionCount.ts
```diff
@@ -0,0 +1,42 @@
+import AsyncStorage from '@react-native-community/async-storage'
+import { useEffect } from 'react'
+
+const SESSION_COUNT_KEY = '@session-count'
+
+const getSessionCount = async () => {
+  const sessionCount = await AsyncStorage.getItem(SESSION_COUNT_KEY)
```
piazzatron, line 7: nit: if this returned 0 when storage was uninitialized rather than null, a lot of the code below could be simplified

src/utils/useSessionCount.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+/**
+ * Invokes `callback` every `frequency` sessions
+ * @param callback
+ * @param frequency
+ * @param startAt which session to start at
+ */
+const useSessionCount = (
```
piazzatron, line 26: It seems like this guy is really sensitive to a non-memoized callback being passed in - if we're chilling on a % 0 and the callback changes once per render or something, I believe we'll call `work` multiple times? Maybe we need to track some state so that we only call work once per start value?

src/components/notification-reminder/NotificationReminder.tsx
```diff
@@ -0,0 +1,59 @@
...
+            // Some subset of notification features are enabled, let the user be.
+            // Do nothing.
+            return
+          case RESULTS.GRANTED:
+            // The user already has given notifications permissions!
+            // Do nothing.
+            return
+          case RESULTS.BLOCKED:
+            // The user has explicitly blocked notifications.
+            // Don't be a jerk.
```
piazzatron, line 36: good guy audius

piazzatron (approved):

src/utils/useSessionCount.ts
raymondjacobson: good callout.

PR #32, opened 2021-01-13
Fix StatusBar colors

piazzatron:
See: https://github.com/AudiusProject/audius-client/pull/194

Additionally, I now have the StatusBar invisible now during iOS splashscreen (since we don't necessarily know the correct color then, and the splash screen looks better without a StatusBar).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "react-native-static-server": "^0.4.2",
     "react-native-version-number": "^0.3.6",
     "react-native-video": "file:react-native-video-5.1.0-alpha2.tgz",
-    "react-native-webview": "^7.4.3",
+    "react-native-webview": "^11.0.2",
     "react-redux": "^7.1.3",
     "redux": "^4.0.4"
   },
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "react-native-static-server": "^0.4.2",
     "react-native-version-number": "^0.3.6",
     "react-native-video": "file:react-native-video-5.1.0-alpha2.tgz",
-    "react-native-webview": "^7.4.3",
+    "react-native-webview": "^11.0.2",
```
raymondjacobson, line 52: damn 3 majors??!? wewwwwww. hopefully this fixes a bunch of stuff. excited to get this on testflight.

src/components/web/WebApp.tsx
```diff
@@ -445,6 +450,7 @@ const WebApp = ({
         onRefresh={() => setAtTop(false)}
       >
         <WebView
+          autoManageStatusBarEnabled={false}
```
raymondjacobson, line 453: maybe comment what this was for

raymondjacobson (approved):

src/components/web/WebApp.tsx
piazzatron: good idea

package.json
piazzatron: 🤞 this fixes you know what

raymondjacobson: who shall not be named

PR #46, opened 2021-02-27
Fix captcha redirect

raymondjacobson:
We have a URL check so that anytime you hit a link from inside the web app, if it's an audius-related link, it stays inside the web app. Looks like the captcha makes an outgoing behind the scenes page change/request to google, so we should also treat that as something within the app.


vicky-g (approved):
😭  recaptcha y

PR #58, opened 2021-03-27
Allow Twitter embed URLs & Bump version

piazzatron:


raymondjacobson (approved):
Don't need to actually increase the version since we never released 1.0.55. Just the build number needs to change but this also works :)

PR #61, opened 2021-03-31
The one that fixes lint

raymondjacobson:
Can be reviewed in two chunks.

The first commit was a bunch of modifications to the .eslintrc setup, etc. and package installs. The second commit is the one where I actually fixed lint issues that were not lint:fix-able. More attention should be given to the second commit.


piazzatron (approved):
LGTM! thx for doing this.

I don't think we want to check in .eslintcache - I have this gitignored on client.

raymondjacobson:
> LGTM! thx for doing this.
> 
> I don't think we want to check in .eslintcache - I have this gitignored on client.

Thx good catch

PR #73, opened 2021-04-13
Remove iOS react native cache

sddioulde:
tested with simulator and local client build against stage. seems to have worked a couple of times.


package.json
```diff
@@ -53,7 +53,7 @@
     "react-native-music-control": "^0.11.0",
     "react-native-permissions": "^3.0.0",
     "react-native-push-notification": "^3.1.9",
-    "react-native-static-server": "^0.5.0",
+    "react-native-static-server": "file:react-native-static-server.tgz",
```
raymondjacobson, line 56: you'll want to commit the tgz file for this to work, but i think i do have a local repro & would be curious to test w/ this package update

sddioulde: the file is committed indeed :)

src/components/web/WebApp.tsx
```diff
@@ -521,6 +521,7 @@ const WebApp = ({
             'mailto://*'
           ]}
           cacheEnabled={false}
+          cacheMode='LOAD_NO_CACHE'
```
sddioulde, line 524: I don't think we need this one. Seems to be android-only related https://github.com/react-native-webview/react-native-webview/blob/404e3e69e23d8c73bc4deb688c77b9a1a0bb55f8/docs/Reference.md#cachemode, unless we have similar issue with android.

raymondjacobson (approved):
Big diff b/c of the lockfiles, but overall LGTM. Will want to QA with intent

PR #77, opened 2021-04-23
Mobile Search

jowlee:
## Summary
Adds native search to mobile

**Changes**
* Adds 2 packages for masked view and gradient that enables text gradient
* Adds the webRef into a context so that it can be used more easily in hooks
* Adds the autocomplete search page & store


src/components/search/Search.tsx
```diff
@@ -0,0 +1,157 @@
...
+    }).start(() => {
+      setDidOpen(true)
+    })
+  }, [fadeAnim, setViewDisplay, setDidOpen])
+
+  const fadeOut = useCallback(() => {
+    Animated.timing(fadeAnim, {
+      toValue: 0,
+      duration: FADE_DURATION,
+      useNativeDriver: true
```
raymondjacobson, line 79: yessss

src/components/search/TopBar.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const query = useSelector(getSearchQuery)
+  const dispatch = useDispatch()
+  const setQuery = useCallback((text: string) => {
+    dispatch(updateQuery(text))
+  }, [dispatch])
+
+  const onSubmit = useCallback(({ nativeEvent: { text } }) => {
+    appendSearchItem(text)
+    dispatchWeb({
+      type: MessageType.UPDATE_SEARCH_QUERY,
```
raymondjacobson, line 109: only real comment is why have update in addition to just a single search that replaces. seems like the action on the web side just does the fetch

raymondjacobson (approved):

src/components/search/TopBar.tsx
jowlee: Yeah, I guess we could just have one action

PR #86, opened 2021-05-03
Search edits

jowlee:
## Summary
* Fix text shadow on the header
* Prevent the updated search text from appearing in history before the results appear

### DIFF
--- a/src/components/search/SearchHistory.tsx
+++ b/src/components/search/SearchHistory.tsx
@@ -114,12 +114,11 @@ const SearchHistoryItem = ({ text }: SearchHistoryItemProps) => {
 
 const SearchHistory = () => {
   const { searchHistory, clearHistory, hasFetched } = useSearchHistory()
-  const [displayHistory, setDisplayHistory] = useState(searchHistory)
+  const [displayHistory, setDisplayHistory] = useState(searchHistory.current)
 
   // Only update history on component load and clearing of search history
   useEffect(() => {
-    if (hasFetched) setDisplayHistory(searchHistory)
-    // @ts-ignore do not reset on searchHistory update.
+    if (hasFetched) setDisplayHistory(searchHistory.current)
   }, [hasFetched, setDisplayHistory])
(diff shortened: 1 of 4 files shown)


src/components/search/SearchHistory.tsx
```diff
@@ -113,31 +113,43 @@ const SearchHistoryItem = ({ text }: SearchHistoryItemProps) => {
...
 const SearchHistory = () => {
-  const { searchHistory, clearHistory } = useSearchHistory()
+  const { searchHistory, clearHistory, hasFetched } = useSearchHistory()
+  const [displayHistory, setDisplayHistory] = useState(searchHistory)
+
+  // Only update history on component load and clearing of search history
+  useEffect(() => {
+    if (hasFetched) setDisplayHistory(searchHistory)
+    // @ts-ignore do not reset on searchHistory update.
+  }, [hasFetched, setDisplayHistory])
```
raymondjacobson, line 123: could be error prone. would a ref make more sense?

raymondjacobson (approved):

src/components/search/SearchHistory.tsx
jowlee: The search results that are displayed are only updated when the history has been fetch and the "clear" button is hit. 
Otherwise, the component is unmounted and re-fetches the history on component load. 
I don't think a ref would cause a re-render of the search history when hitting the clear btn.

raymondjacobson: you could have both a state var and a ref

jowlee: Are you suggesting that the update flow stay the same w/ this use effect and update in onClearHistory, just changing the `displayHistory` to be a ref instead of state?

raymondjacobson: searchHistory so that you don't have to ts ignore that's all

raymondjacobson:
thx :)

PR #99, opened 2021-08-11
Update readme and firebase deps

raymondjacobson:
### Description

Makes it so we don't have to do any extra work on M1 to get the iOS simulator working. (esp between pod installs)

On M1, you should do 
```
brew install ruby
sudo gem install cocoapods
```
so pod install works

### Dragons

Is there anything the reviewer should be on the lookout for? Are there any dangerous changes?

Some package version upgrades. Will want to look closely, but just coming from a `Pod update`

### How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration.

Tested on M1 and intel-based

For features that are critical or could fail silently please describe the monitoring/alerting being added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,40 @@
 # Audius Mobile Client
 
 The Audius React Native mobile client.
-It can be run against a local client or against a production build.
+
+This project is a React Native wrapper around the Audius web client, and requires a web client to be running.
+The native project can be built & run against a local client (serving at localhost) or against a vendored staging/production build.
 
 ## Setup
 
-Copy the environment variables and replace missing values (you will need an FCM sender id as well as a Segment write key)
+Copy the environment variables and replace missing values. (You will need an FCM sender id as well as a Segment write key for those services to work properly, but any value will suffice if the data is not important to you.)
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -62,15 +90,6 @@ npm run copy:remote-production
...
-# install local dependencies
-npm install
-
-cd ios
-pod install
-cd ..
-
 # Run a simulator pointed at a static build
 npm run ios
 # Run a simulator pointed at localhost
```
sliptype, line 95:
```suggestion
# Run a simulator pointed at URL_OVERRIDE
```

sliptype: I think we should revisit the comments in this section because `npm run ios` vs `npm run ios:dev` isn't necessarily static build vs localhost. It's just using different .env files

sliptype (approved):
This is awesome!

Pointed out a small thing I think we could make clearer still

Once this is in I can deprecate that M1 doc too

raymondjacobson:
Will merge this PR after tik tok launch as to not complicate things

PR #110, opened 2021-09-03
Add notifications for challenge rewards

Kyle-Shanks:
### Description

Added code to display notifications for challenge rewards

### How Has This Been Tested?

Tested in mobile emulator

### DIFF
--- a/src/components/notifications/content/ChallengeReward.tsx
+++ b/src/components/notifications/content/ChallengeReward.tsx
@@ -1,0 +1,89 @@
+import React from 'react'
+import { StyleSheet, Text, View } from 'react-native'
+import { useTheme } from '../../../utils/theme'
+import {
+  ChallengeReward as ChallengeRewardType,
+  ChallengeRewardID
+} from 'store/notifications/types'
+import TwitterShare from './TwitterShare'
+
+const styles = StyleSheet.create({
+  wrapper: {
+    display: 'flex',
+    flexDirection: 'column'
+  },
+  text: {
+    fontFamily: 'AvenirNextLTPro-Bold',
+    fontSize: 16,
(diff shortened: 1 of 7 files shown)


src/components/notifications/content/ChallengeReward.tsx
```diff
@@ -0,0 +1,89 @@
...
+type ChallengeRewardProps = {
+  notification: ChallengeRewardType
+  onGoToRoute: (route: string) => void
+}
+
+const challengeInfoMap: Record<
+  ChallengeRewardID,
+  { title: string; amount: number }
+> = {
+  'profile-completion': {
```
raymondjacobson, line 31: i wonder if we can/should pull these in via optimizely. e.g. if we change the 7 days to 6 days or something

raymondjacobson (approved):

src/components/notifications/content/ChallengeReward.tsx
Kyle-Shanks: 💯 I agree. This same type of config is in 3 places now so I want to put it somewhere central for them all to read from instead

raymondjacobson: maybe the client could fetch from discovery

raymondjacobson: the more i think about it, the more i like this. wdyt?

raymondjacobson: also down to merge this as is first

Kyle-Shanks: Cool will merge now and refactor in another PR. And yea I'd like to discuss that approach 👍

PR #112, opened 2021-09-10
AUD-673 Minor message passing cleanup

sliptype:
### Description

This PR:
* Splits out the message handlers into separate files

Future scope:
* Share message types between client and mobile repos
* Rethink "message passing" in general, perhaps create an abstraction over it using the redux api so the mobile client can transparently interact with the client store

### Dragons

This touches all message handlers, I ran through them and everything still seems to be working. Also added a console warn if a message is received without a matching handler

### How Has This Been Tested?


src/components/web/WebApp.tsx
```diff
@@ -565,10 +565,9 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
   onMessage: (
     message: Message,
     onPostMessage: (message: Message) => void,
-    reload: () => void,
-    state: AppState
+    reload: () => void
   ) => {
-    handleMessage(message, dispatch, onPostMessage, reload, state)
+    handleMessage(message, dispatch, onPostMessage, reload)
```
raymondjacobson, line 570: nice. wonder why we were passing in state before lol

src/message/handleMessage.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const handleMessage = async (
+  message: Message | AnalyticsMessage,
+  dispatch: Dispatch,
+  postMessage: (message: Message) => void,
+  reload: () => void
+) => {
+  const handler = messageHandlers[message.type]
+  if (handler) {
+    handler(message, dispatch, postMessage, reload)
```
raymondjacobson, line 47: I think this would be a bit nicer if we did 

{ message, dispatch, postMessage, reload }

instead of position args. would remove a lot of the `_` right?

Also, might be worth adding a quick comment here on what these things do (e.g. that `postMessage` is your way to respond back)

raymondjacobson (approved):
love it

src/message/handleMessage.ts
sliptype: Agreed, I'm a big fan of the `options` argument pattern on anything over arity 3. Might add that to the js styleguide at some point

raymondjacobson: i think even at 1 thing the options arg is better. so much more extensible

PR #142, opened 2021-11-04
Fix demi bold font

raymondjacobson:
### Description

This was really rough.

TL;DR "demi" is not a valid weight suffix, but "demibold" is
https://github.com/facebook/react-native/blob/main/React/Views/RCTFont.mm#L23

Discussion:
https://github.com/facebook/react-native/issues/20788

### Dragons

### How Has This Been Tested?

Locally w/ sign up flow


sliptype (approved):
Interesting

I think we should add DemiBold to the `Text` component as well

PR #144, opened 2021-11-05
Fix notif drawer and bump build

raymondjacobson:
### Description

Make notif drawer show on sign up not just sign in

### Dragons

### How Has This Been Tested?


src/components/signon/SignOn.tsx
```diff
@@ -366,11 +366,9 @@ const SignOn = ({ navigation }: SignOnProps) => {
       setEmail('')
       setPassword('')
```
sliptype, line 368: Not sure if `signedIn` is true after sign up but I assume it is

sliptype (approved):

src/components/signon/SignOn.tsx
sddioulde: it is

PR #145, opened 2021-11-05
Fix additional mobile signon bugs

sddioulde:
### Description

Fix additional mobile signon bugs

### Dragons

### How Has This Been Tested?

Locally spin up ios and android emulators

### DIFF
--- a/src/components/notifications/Notifications.tsx
+++ b/src/components/notifications/Notifications.tsx
@@ -291,7 +291,7 @@ const Notifications = ({
 
   const panProps = canShowNotifications ? panResponder.panHandlers : {}
 
-  return (
+  return isSignedIn ? (
     <>
       {/* animated drawer */}
       <Animated.View
@@ -328,7 +328,7 @@ const Notifications = ({
         ]}
       />
     </>
-  )
+  ) : null
 }
 
 const mapStateToProps = (state: AppState) => ({
--- a/src/components/signon/SignOn.tsx
+++ b/src/components/signon/SignOn.tsx
@@ -18,6 +18,7 @@ import {
 } from 'react-native'
 import RadialGradient from 'react-native-radial-gradient'
 import { useDispatch, useSelector } from 'react-redux'
+import { getIsKeyboardOpen } from '../../store/keyboard/selectors'
 import backgImage from '../../assets/images/DJportrait.jpg'
 import audiusLogoHorizontal from '../../assets/images/Horizontal-Logo-Full-Color.png'
 import IconArrow from '../../assets/images/iconArrow.svg'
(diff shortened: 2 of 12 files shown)


src/components/signon/SignOn.tsx
```diff
@@ -321,6 +324,7 @@ const SignOn = ({ navigation }: SignOnProps) => {
   const emailIsAvailable = useSelector(getEmailIsAvailable)
   const emailIsValid = useSelector(getEmailIsValid)
   const emailStatus = useSelector(getEmailStatus)
+  const isKeyboardVisible = useSelector(getIsKeyboardVisible)
```
sliptype, line 327: I think you can use the existing selector for this:
https://github.com/AudiusProject/audius-mobile-client/blob/master/src/store/keyboard/selectors.ts#L4

And then you won't need to make any changes to `useKeyboardListeners`

sddioulde: Oh nice! Thanks

src/components/notifications/Notifications.tsx
```diff
@@ -291,7 +291,7 @@ const Notifications = ({
 
   const panProps = canShowNotifications ? panResponder.panHandlers : {}
 
-  return (
+  return isSignedIn ? (
```
raymondjacobson, line 294: is this true when the session first loads? (i think probably, but just want to make sure notifs are accessible right away still)

src/store/keyboard/sagas.ts
```diff
@@ -1,11 +1,30 @@
...
+    Keyboard.addListener('keyboardDidShow', () => {
+      emitter('show')
+    })
+    Keyboard.addListener('keyboardDidHide', () => {
+      emitter('hide')
+    })
+    return () => {}
+  })
+
+  yield fork(function* () {
```
raymondjacobson, line 18: what does the fork gain here? probably can just put the while (true) { } in line right?

src/store/sagas.ts
```diff
@@ -1,5 +1,10 @@
...
+import { all, call } from 'redux-saga/effects'
+import initKeyboardEvents from './keyboard/sagas'
 
 export default function* rootSaga() {
-  return [initKeyboardEvents]
+  // executing array of effects was deprecated in v1 of redux-saga
+  // must call them in the all effect now
+  // https://stackoverflow.com/questions/55811145/redux-saga-does-not-run/55811902#55811902
+  const sagas = [initKeyboardEvents]
+  yield all(sagas.map(call))
```
raymondjacobson, line 9: https://github.com/AudiusProject/audius-client/blob/master/src/store/sagas.js#L149

raymondjacobson (approved):

src/store/signon/reducer.ts
```diff
@@ -39,6 +39,7 @@ export type SignonState = {
   }
   finalEmail: string
   finalHandle: string
+  isKeyboardVisible: boolean
```
sliptype, line 42: This can be removed now

src/components/notifications/Notifications.tsx
sddioulde: yes it is

PR #148, opened 2021-11-09
[AUD-1014] ActionDrawer

sliptype:
### Description

This PR:
* Implements the `ActionDrawer` and `OverflowMenuDrawer`
Simulator Screen Shot - iPhone 11 - 2021-11-10 at 14 37 09

### Dragons

### How Has This Been Tested?


raymondjacobson:
random drive by --- in the screenshot, i think there should be no line on the top of "Repost"

sliptype:
> random drive by --- in the screenshot, i think there should be no line on the top of "Repost"

Whoops, you're right. For some reason I had it in my head that the border should be there, I even wrote extra styles for it 🤦

Fixed now

src/components/notifications/Notifications.tsx
```diff
@@ -142,17 +141,6 @@ const Notifications = ({
     markAsViewed()
   }, [initialPosition, close, markAsViewed, backgroundAnim, translationAnim])
 
-  useAppState(
```
sliptype, line 145: @raymondjacobson After messing with entering/exiting background with the notifications open, I couldn't find a case were this was needed. So removing it for now, let's keep and eye on this during testing though

src/components/overflow-menu-drawer/CollectionOverflowMenuDrawer.tsx
```diff
@@ -0,0 +1,88 @@
...
+    playlist_name,
+    is_album
+  } = useSelectorWeb((state: CommonState) => getCollection(state, { id: id }))
+
+  const { handle } = useSelectorWeb((state: CommonState) =>
+    getUser(state, { id: playlist_owner_id })
+  )
+
+  if (!id || !handle || !playlist_name || is_album === undefined) {
+    return <></>
```
raymondjacobson, line 59: why fragment instead of null?

src/components/overflow-menu-drawer/TrackOverflowMenuDrawer.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const { handle } = useSelectorWeb((state: CommonState) =>
+    getUser(state, { id: owner_id })
+  )
+
+  if (!id || !owner_id || !handle || !title) {
+    return <></>
+  }
+
+  const callbacks = {
```
raymondjacobson, line 60: i like this pattern. is it worth a useMemo?

raymondjacobson (approved):

src/components/overflow-menu-drawer/CollectionOverflowMenuDrawer.tsx
sliptype: Because I neglected the styleguide

 Fixed!

src/components/overflow-menu-drawer/TrackOverflowMenuDrawer.tsx
sliptype: I don't think it's worth it, because this is near the top of the React tree this should only ever rerender when the `mobileOverflowModal` state changes. In which case these callbacks need to be redefined anyway

PR #149, opened 2021-11-09
Fix autoplay app crash bug

sddioulde:
### Description

App used to crash when on a lineup that has only one track. We were trying to pull the genre off a non-existent track, which caused an exception.

This PR also handles triggering autoplay from native mobile when close to end of lineup.

### Dragons

### How Has This Been Tested?

iPhone and Android emulators

### DIFF
--- a/src/components/audio/Audio.tsx
+++ b/src/components/audio/Audio.tsx
@@ -365,7 +365,7 @@ const Audio = ({
       isCloseToEndOfQueue
     ) {
       postMessage(webRef.current, {
-        type: MessageType.QUEUE_AUTOPLAY,
+        type: MessageType.REQUEST_QUEUE_AUTOPLAY,
         genre: (track && track.genre) || undefined,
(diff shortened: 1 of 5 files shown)


src/components/audio/Audio.tsx
```diff
@@ -343,6 +344,41 @@ const Audio = ({
...
+    const isNotRepeating = repeatMode === RepeatMode.OFF
+    if (
+      webRef.current &&
+      queueAutoplay &&
+      !isShuffleOn &&
+      isNotRepeating &&
+      isCloseToEndOfQueue
+    ) {
+      postMessage(webRef.current, {
+        type: MessageType.QUEUE_AUTOPLAY,
```
raymondjacobson, line 363: maybe we could make this action verb prefixed so it's clear what we want.. something like

FETCH_QUEUE_AUTOPLAY or REQUEST_QUEUE_AUTOPLAY or something like that

raymondjacobson (approved):

PR #154, opened 2021-11-18
Adds amplitude inplace of segment

jowlee:
### Description
Removes segment and adds amplitude.

Note: the CircleCI env vars hav been added

### How Has This Been Tested?
I ran the mobile client locally with an ios and android emulator and checked that the events were on amplitude's dashboard

amplitude logs


package.json
```diff
@@ -36,6 +36,7 @@
     "prepare": "husky install"
   },
   "dependencies": {
+    "@amplitude/react-native": "^2.6.0",
```
sliptype, line 39: 📌
```suggestion
    "@amplitude/react-native": "2.6.0",
```

src/types/analytics.ts
```diff
@@ -50,7 +50,10 @@ export enum EventNames {
   LISTEN = 'Listen',
 
   // Notifications
-  NOTIFICATIONS_OPEN_PUSH_NOTIFICATION = 'Notifications: Open Push Notification'
+  NOTIFICATIONS_OPEN_PUSH_NOTIFICATION = 'Notifications: Open Push Notification',
+
+  // Notifications
```
sliptype, line 55: This comment is duplicated

sliptype (approved):
Nice!

raymondjacobson:
Let's hold off on merge until we ship the next round of mobile updates, which have a lot of stuff!

.env.dev.tmpl
```diff
@@ -5,6 +5,5 @@ AUDIUS_URL=https://audius.co
 USER_NODE=https://usermetadata.audius.co
 LEGACY_USER_NODE=https://creatornode.audius.co
 IDENTITY_SERVICE=https://identityservice.staging.audius.co
-SEGMENT_IOS_WRITE_KEY=abc
-SEGMENT_ANDROID_WRITE_KEY=abc
+AMPLITUDE_WRITE_KEY=abc
```
rickyrombo, line 8: @jowlee I ran into an annoying crash issue because `.stage.env` was missing the `SEGMENT_IOS_WRITE_KEY` which I was gonna put a PR up to add to the template, but since you're replacing it here can you add `AMPLITUDE_WRITE_KEY` to the `.stage.env.tmpl` (if this has the same problem?)

rickyrombo: Maybe prod too, as long as it doesn't interfere with how those envs are set for actual deployments

sliptype:
Oh do you need to update the circleci config to copy the correct env vars? 

These lines:
```
echo "SEGMENT_IOS_WRITE_KEY=$SEGMENT_IOS_WRITE_KEY" >> .env.prod
```

PR #155, opened 2021-11-19
[AUD-1065] Top level stack navigation setup

sliptype:
### Description

This PR is a no-op as long as the `IS_MAIN_NAVIGATION_ENABLED` flag is set to false. Once it is set to true the RN bottom bar and navigation will be shown

* Adds an example Feed stack
* Integrates the sign on stack into the app navigator
* Allows different bottom tab routes to go to RN screens & webview screens
* Configures basic deep linking

### Dragons

Moves the NavigationContainer from the sign on navigator to the top app level, so I tested to make sure all sign on flows are still working

### How Has This Been Tested?

Tested on ios & android device + ios simulator


src/components/animated-button/AnimatedButtonProvider.tsx
```diff
@@ -22,44 +29,53 @@ type AnimatedButtonProps = {
 
 const AnimatedButton = ({
   iconJSON,
-  onClick,
+  onPress,
```
raymondjacobson, line 32: ty

src/components/animated-button/AnimatedButtonProvider.tsx
```diff
@@ -70,8 +86,8 @@ const AnimatedButton = ({
 
 export type AnimatedButtonProviderProps = {
   isDarkMode: boolean
-  iconDarkJSON: () => any
-  iconLightJSON: () => any
+  iconDarkJSON: any
```
raymondjacobson, line 89: probably I did this, but the any was necessary b/c the JSON wasn't well typed. maybe we could at least do

`type JSON = any` or (object?) and reuse that here or something

src/components/notifications/Notifications.tsx
```diff
@@ -106,11 +110,21 @@ const Notifications = ({
   }, [webRef])
 
   const { width } = Dimensions.get('window')
-  const initialPosition = -1 * width + INITIAL_OFFSET
+
+  // When a navigation stack is open, don't allow the notifications to be opened
```
raymondjacobson, line 114: this seems find, but is a bit wonky (the initial way i set it up was wonky). I wonder if there is a more canonical navigation stack solution so that the left swipe gesture is *fully* supported rather than there just being this slight border we can latch onto

src/components/signon/types.ts
```diff
@@ -0,0 +1,20 @@
+export type SignOnStackParamList = {
+  SignOn: undefined
```
raymondjacobson, line 2: what is this type

src/screens/profile-screen/ProfileScreen.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const ProfileScreen = ({ navigation }: Props) => {
+  return (
+    <View>
+      <Text>Example profile screen</Text>
+    </View>
+  )
+}
+
+export default ProfileScreen
```
raymondjacobson, line 20: I think we're missing an index.ts exporter file in this folder

raymondjacobson (approved):
This all looks fantastic to me.
One question about what is a "screen" vs, a "component"

Is the idea that the screen just houses the top level display logic, but the components themselves that make up the page live in components/?

src/components/notifications/Notifications.tsx
sliptype: I think we could probably use react gesture handler or something to catch a swipe from the left edge

src/components/signon/types.ts
sliptype: This defines the types of the params for each screen in the navigator. `undefined` means there are no params for that screen

src/components/app-navigator/BottomTabNavigator.tsx
```diff
@@ -65,30 +66,67 @@ const FeedStackScreen = createStackScreen<FeedStackParamList>(Stack => (
...
-        <Tab.Screen name='favorites' component={EmptyScreen} />
-        <Tab.Screen name='profile' component={EmptyScreen} />
-      </Tab.Navigator>
-    </View>
+    <>
+      <View style={styles.tabNavigator}>
+        <Tab.Navigator
+          tabBar={props => (
+            <>
+              <NowPlayingDrawer
```
sliptype, line 106: I had to move the NowPlayingDrawer into the "tabBar" of the Tab.Navigator, because there was no way to get it to sit between the screen and the tab bar otherwise

src/components/app-navigator/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,132 @@
...
+ * The bottom tab navigator
+ *
+ * TODO: This navigator is only displayed when the user is authed, so we should
+ * move drawers, modals, notifications in here. Need to wait until fully migrated to RN
+ * because of the way the top level navigator is hidden to display the WebView
+ */
+const BottomTabNavigator = ({
+  onBottomTabBarLayout
+}: BottomTabNavigatorProps) => {
+  // Set handlers for the NowPlayingDrawer and BottomTabBar
```
raymondjacobson, line 83: These are way better here, thanks

raymondjacobson (approved):

PR #176, opened 2021-12-03
[AUD-1011] Add connect wallets drawer

Kyle-Shanks:
### Description
Add connect wallets drawer

### How Has This Been Tested?

Manually tested. It looks great

### DIFF
--- a/src/components/connect-wallets-drawer/ConnectWalletsDrawer.tsx
+++ b/src/components/connect-wallets-drawer/ConnectWalletsDrawer.tsx
@@ -1,0 +1,84 @@
+import React, { useCallback } from 'react'
+
+import MaskedView from '@react-native-masked-view/masked-view'
+import {
+  getModalVisibility,
+  setVisibility
+} from 'audius-client/src/common/store/ui/modals/slice'
+import { StyleSheet, View } from 'react-native'
+import LinearGradient from 'react-native-linear-gradient'
+
+import Drawer from 'app/components/drawer'
+import Text from 'app/components/text'
+import { useDispatchWeb } from 'app/hooks/useDispatchWeb'
+import { useSelectorWeb } from 'app/hooks/useSelectorWeb'
+import { useThemedStyles } from 'app/hooks/useThemedStyles'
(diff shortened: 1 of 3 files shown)


src/components/connect-wallets-drawer/ConnectWalletsDrawer.tsx
```diff
@@ -0,0 +1,86 @@
...
+    text: {
+      fontFamily: 'AvenirNextLTPro-Regular',
+      textAlign: 'center',
+      fontWeight: '500',
+      fontSize: 24,
+      lineHeight: 30
+    }
+  })
+
+// `ActionDrawer` is a drawer that presents a list of clickable rows with text
```
sliptype, line 52: Extra comment

src/components/connect-wallets-drawer/ConnectWalletsDrawer.tsx
```diff
@@ -0,0 +1,86 @@
...
+  StyleSheet.create({
+    container: {
+      paddingTop: 48,
+      paddingLeft: 16,
+      paddingRight: 16,
+      paddingBottom: 48
+    },
+
+    title: {
+      fontFamily: 'AvenirNextLTPro-Heavy',
```
sliptype, line 36: No need to change but you can set `weight="heavy"` on our `Text` component. I think it's a little cleaner than typing out the font family everywhere

sliptype (approved):
Nice!

src/components/connect-wallets-drawer/ConnectWalletsDrawer.tsx
Kyle-Shanks: I agree. Used font weight so I don't have to write it twice though 👍

sliptype: Font-weight doesn't work on Android. Or iOS? one of the two doesn't work lol

Also I think eventually we should turn the masked text thing into a component so we don't have all this duplication everywhere

PR #180, opened 2021-12-10
Ms mobile download

milindNshah:
### Description
When the download button is clicked on a track (from audius-client), trigger the download and save the file onto the user's device. 

1. Add some permission to allow downloads to be triggered from iOS and Android
2. Create a new message handler for "Download-track"
3. Create a new download track drawer
3. Inside message handler
    - Trigger the download drawer to pop up (see designs)
    - Download the file using RNFetchBlob
    - While downloading send updates to % of download completed (on iOS)
    - Once download is completed, hide the download drawer
 4. In case of failure or user cancelling the download, delete the files 

Designs
- https://www.figma.com/file/ii6uNAdGGoQ7X7p7cENGhf/Mobile-Download-Buttons?node-id=101%3A141

### Dragons
- Will be easier to review by commit
- Auto-generated changes in ios/AudiusReactNative.xcodeproj/project.pbxproj
- Auto-generated changes in ios/Podfile.lock
- Currently Android doesn't get updates on progress for download so I removed the numerical percentage from the drawer (issue with RNFetchBlob library)

### How Has This Been Tested?
Download drawer
Screen Shot 2021-12-09 at 10 55 50 AM

Flow 1) User completes download
| iOS | Android |
|--|--|
|https://user-images.githubusercontent.com/12490332/145495958-752b00ce-49a0-40de-90bb-7b2bd573fe00.mp4 | https://user-images.githubusercontent.com/12490332/145495749-72199df2-8f55-40b6-be0a-e05f56851471.mp4|
|||

Flow 2) User cancels download
| iOS | Android |
| -- | -- |
| Cancel error won't show up on prod btw it shows up in video just cause I'm running in debug mode: 
https://user-images.githubusercontent.com/12490332/145496070-45682f5c-c868-4315-85d8-43a21f8573b9.mp4 | https://user-images.githubusercontent.com/12490332/145496374-91530077-19c5-43c3-a322-abae876f0a2f.mp4 |


package.json
```diff
@@ -86,7 +86,8 @@
     "react-redux": "7.2.5",
     "redux": "4.1.1",
     "redux-saga": "1.1.3",
-    "reselect": "4.0.0"
+    "reselect": "4.0.0",
+    "rn-fetch-blob": "^0.12.0"
```
sliptype, line 90: Let's pin this dependency

src/components/download-track-progress-drawer/DownloadTrackProgressDrawer.tsx
```diff
@@ -0,0 +1,110 @@
...
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'center',
+    paddingTop: 16,
+    paddingHorizontal: 24
+  },
+
+  title: {
+    fontFamily: 'AvenirNextLTPro-Bold',
+    fontSize: 21,
```
sliptype, line 27: Are these colors all the same for dark mode? I would recommend using the `useThemedStyles` hook to get access to the theme colors

src/components/download-track-progress-drawer/DownloadTrackProgressDrawer.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  loadingIcon: {
+    width: 40,
+    height: 40,
+    alignSelf: 'flex-start',
+    marginRight: 10
+  },
+
+  downloadPercent: {
+    fontFamily: 'AvenirNextLTPro-Heavy',
```
sliptype, line 56: More of a preference but if you use our `Text` component you can specify the weight prop and don't have to type out the font family everywhere

src/components/download-track-progress-drawer/DownloadTrackProgressDrawer.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (fetchTask) {
+      fetchTask.cancel()
+    }
+    setIsOpen(false)
+  }, [fetchTask, setIsOpen])
+
+  return (
+    <Drawer isOpen={isOpen} onClose={handleClose}>
+      <View style={styles.view}>
+        <Text style={styles.title}>Downloading</Text>
```
sliptype, line 80: Can put this text in a messages object? That way it'll will be easier if we support i18n in the future

src/store/download/slice.ts
```diff
@@ -0,0 +1,52 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import { FetchBlobResponse, StatefulPromise } from 'rn-fetch-blob'
+
+export type DownloadState = typeof initialState
+
+type State = {
+  downloadedPercentage: number
+  fetchTask: StatefulPromise<FetchBlobResponse>
```
sliptype, line 8: Generally only plain js objects should be stored in redux. This is to allow for things like time travel debugging and serializing the state. Is there any way you could store this fetch task elsewhere? Could just be a mutable variable in the handler file I think

sliptype (approved):
🔥🔥🔥 looks great!

Left some comments

src/store/download/slice.ts
milindNshah: Hmm so the problem is that it needs to be accessed by the DownloadTrackProgressDrawer. I don't think putting it in the handler files would let DownloadTrackProgressDrawer access it. And I don't think we can pass the fetchTask object to the DownloadTrackProgressDrawer any other way. The only alternative I can think of, without putting it in the redux store is to make it a globally accessible variable but that doesn't seem like a great choice either. Happy to discuss further!

src/components/download-track-progress-drawer/DownloadTrackProgressDrawer.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <View style={styles.view}>
+        <Text style={styles.title} weight='bold'>
+          {messages.title}
+        </Text>
+        <Text style={styles.fileName} weight='bold'>
+          {fileName}
+        </Text>
+        <View style={styles.downloadContainer}>
+          <LottieView
+            style={styles.loadingIcon}
```
raymondjacobson, line 96: probably worth extracting into a shared component for re-use. no action here but figured i'd flag.

src/store/download/slice.ts
raymondjacobson: an alternative here that would be more canonical would be to expose another action that the component could call to find & cancel the ongoing fetch task. that would let you track the fetch task entirely within the message handler/saga layer and not bleed into the store/component. 

A comment here about why we're doing this would also be reasonable, but i don't think this change would be too hard.

raymondjacobson (approved):

src/store/download/slice.ts
milindNshah: Updated to pass the "cancel function" instead of the stateful promise!

PR #184, opened 2021-12-16
[AUD-1004] Trending Rewards Explainer Drawer

sliptype:
### Description

This PR:
* Adds `TrendingRewardsDrawer`
Simulator Screen Shot - iPhone 13 - 2021-12-16 at 13 58 09
* Adds  the following components:
  * `TabSlider`
  * `GradientText`
  * `LoadingSpinner`
  * `TweetEmbed`
  * `ButtonWithArrow`
* Downgrades `redux-saga` to match the client. This is needed to share sagas via common
* Adds remote config state and updates corresponding hooks

### Dragons

The corresponding client PR needs to be merged and the sha updated https://github.com/AudiusProject/audius-client/pull/699

### How Has This Been Tested?

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "react-native-webview": "11.13.1",
     "react-redux": "7.2.5",
     "redux": "4.1.1",
-    "redux-saga": "1.1.3",
+    "redux-saga": "0.16.2",
     "reselect": "4.0.0",
     "rn-fetch-blob": "0.12.0"
   },
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@react-navigation/stack": "6.0.11",
     "@reduxjs/toolkit": "1.6.1",
     "@svgr/core": "5.5.0",
-    "audius-client": "github:AudiusProject/audius-client.git#f34b60dfe87e3eda09a9b17ca44b13f7cae5ae25",
+    "audius-client": "github:AudiusProject/audius-client.git#2bd0f8c8fbb79f318487d8c748f709200fc8546e",
     "bn.js": "5.2.0",
     "expo-crypto": "9.2.0",
     "fxa-common-password-list": "0.0.4",
--- a/src/components/button/Button.tsx
+++ b/src/components/button/Button.tsx
@@ -7,7 +7,8 @@ import {
   StyleSheet,
   TextStyle,
   View,
-  Animated
+  Animated,
+  StyleProp
 } from 'react-native'
 
 import Text from 'app/components/text'
@@ -73,7 +74,7 @@ const createStyles = (type: ButtonType = ButtonType.PRIMARY) => (
           fontFamily: 'AvenirNextLTPro-Bold'
         },
         icon: {
-          marginLeft: 12
+          marginLeft: 6
         },
         disabled: {
           backgroundColor: '#E7E6EB'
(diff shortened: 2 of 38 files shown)


src/components/button/Button.tsx
```diff
@@ -73,7 +74,7 @@ const createStyles = (type: ButtonType = ButtonType.PRIMARY) => (
           fontFamily: 'AvenirNextLTPro-Bold'
         },
         icon: {
-          marginLeft: 12
+          marginLeft: 6
```
milindNshah, line 77: This will change the margin for every button - do we want that?

src/components/tweet-embed/TweetEmbed.tsx
```diff
@@ -46,23 +46,25 @@ const TweetEmbed = ({ options, tweetId }: Props) => {
...
-    </View>
+      <View style={{ width: '100%', height }}>
+        <WebView
+          onMessage={event => {
+            setHeight(parseInt(event.nativeEvent.data, 10))
+            setIsLoading(false)
+          }}
+          scrollEnabled={false}
+          source={{
+            html
```
milindNshah, line 72: Can we move the html code into here so it's not inside a string?

package.json
```diff
@@ -87,7 +87,7 @@
     "react-native-webview": "11.13.1",
     "react-redux": "7.2.5",
     "redux": "4.1.1",
-    "redux-saga": "1.1.3",
+    "redux-saga": "0.16.2",
```
milindNshah, line 90: Any side effects to watch out for by reverting this verison?

src/components/drawer/Drawer.tsx
```diff
@@ -534,6 +534,7 @@ const Drawer = ({
       >
         <SafeAreaView
           edges={['bottom', ...((isFullscreen ? ['top'] : []) as Edge[])]}
+          style={{ height: isFullscreen ? '100%' : 'auto' }}
```
milindNshah, line 537: Nit: I think it might be better practice to have two different classes declared within createStyles and pick one based on the isFullScreen boolean so as to keep all styling in one place and not spread out throughout the file

milindNshah (approved):
Left some minor comments below, otherwise LGTM!

src/components/gradient-text/GradientText.tsx
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * The text to be displayed
+   */
+  text: string
+}
+
+/**
+ * Diagonal gradient text in the Audius colors
+ */
+const GradientText = ({ style, text }: Props) => {
```
Kyle-Shanks, line 24: NICE 🔥

src/components/button/Button.tsx
Kyle-Shanks: ^

Kyle-Shanks (approved):
Looks good, love the GradientText and LoadingSpinner additions

src/components/tweet-embed/TweetEmbed.tsx
sliptype: Unfortunately react-native-webview only accepts a string for the `html` prop https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#source

We have to do something similar (albeit with js code) in https://github.com/AudiusProject/audius-mobile-client/blob/master/src/components/oauth/OAuth.tsx because the metro bundler doesn't support loading html/js files as a string. Maybe something like this could be possible using handlebars?

package.json
sliptype: I needed to make a small change at the root saga, but we only recently added `redux-saga` to this repo. There is only one other saga, the keyboard open/close state. I tested that it's still working!

src/components/button/Button.tsx
sliptype: I reverted this, I thought 6px looked better across the board but for consistency with designs and web it should be 12px

src/components/drawer/Drawer.tsx
sliptype: I like it! Updated

PR #197, opened 2021-12-23
[AUD-1037]Get filename from client instead of generating it

milindNshah:
### Description
QOL fix, get passed the filename instead of trying to generate it. (Filename is more important than track title, title may not exist in the case of stems)

### How Has This Been Tested?
Tested on iOS simulator, iOS Device, android Device

### DIFF
--- a/src/message/handlers/download.ts
+++ b/src/message/handlers/download.ts
@@ -22,11 +22,11 @@ const cancelDownloadTask = () => {
 export const messageHandlers: Partial<MessageHandlers> = {
   [MessageType.DOWNLOAD_TRACK]: async ({ message }) => {
     const fileUrl = message.urls.find(url => url !== null && url !== undefined)
-    const fileExtension = fileUrl.split('.').pop()
-    const fileName = message.title + '.' + fileExtension
+    const fileName = message.filename
+    const trackName = fileName.split('.').slice(0, -1).join('')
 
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
Not sure why there are so many lock file changes -- maybe we should revert

can this PR also update the audius-client SHA in package.json to pull in the other change you made?

src/message/handlers/download.ts
```diff
@@ -21,11 +21,11 @@ const cancelDownloadTask = () => {
...
   [MessageType.DOWNLOAD_TRACK]: async ({ message }) => {
     const fileUrl = message.urls.find(url => url !== null && url !== undefined)
-    const fileExtension = fileUrl.split('.').pop()
-    const fileName = message.title + '.' + fileExtension
+    const fileName = message.filename
+    const trackName = fileName.split('.').slice(0, -1).join('')
 
     dispatch(setVisibility({ drawer: 'DownloadTrackProgress', visible: true }))
-    dispatch(setFileInfo({ trackName: message.title, fileName: fileName }))
+    dispatch(setFileInfo({ trackName: trackName, fileName: fileName }))
```
sliptype, line 28: Same nit as from the other PR
```suggestion
    dispatch(setFileInfo({ trackName, fileName }))
```

sliptype (approved):

PR #198, opened 2021-12-24
[AUD-1015] TikTok sharing

sliptype:
### Description

This PR:
* Handles the TikTok oauth on the RN side, and sets us up to move instagram/twitter to rn in the future
* Adds the `ShareToTikTokDrawer`
* Sets a default color for the `LoadingSpinner` as something other than white to prevent confusion

### Dragons

### How Has This Been Tested?

Tested on both ios and android devices


src/components/oauth/OAuth.tsx
```diff
@@ -178,23 +177,20 @@ const TIKTOK_POLLER = `
...
-type OwnProps = {
+type Props = {
   webRef: RefObject<MessagePostingWebView>
 }
 
-type Props = OwnProps &
-  ReturnType<typeof mapStateToProps> &
-  ReturnType<typeof mapDispatchToProps>
+const OAuth = ({ webRef }: Props) => {
+  const dispatch = useDispatch()
```
raymondjacobson, line 185: ty

src/hooks/useTikTokAuth.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+    handleError: (e: Error) => {
+      track(
+        make({
+          eventName: EventNames.TIKTOK_OAUTH_ERROR,
+          error: e.message
+        })
+      )
+    },
+    getLocalStorageItem: AsyncStorage.getItem,
```
raymondjacobson, line 41: this pattern is pretty slick. we should adopt it everywhere. cc @jowlee

src/store/oauth/sagas.ts
```diff
@@ -0,0 +1,45 @@
...
+
+function* watchSetCredentials() {
+  const action: SetCredentialsAction = yield take(SET_CREDENTIALS)
+
+  if (!action.credentials.error) {
+    if (resolve) {
+      resolve(action.credentials)
+    }
+  } else {
+    if (reject) {
```
raymondjacobson, line 32: this flow is kind of confusing to me. i think i get it now, but maybe we should have some sort of invariant on resolve & reject being set. what if we put the watchSetCredentials inside the watchRequestNativeOpenPopup?

Also, should these be takeLatest or takeEvery? I think they'd only be allowed to run once the way they're written

raymondjacobson (approved):

src/store/oauth/sagas.ts
sliptype: Ah of course, I forgot how sagas worked for a minute

Good callout, I think this is way cleaner. Is this what you had in mind? https://github.com/AudiusProject/audius-mobile-client/pull/198/commits/f03462eff81a4c1a6ef6220097f70a8a8fe993ea

raymondjacobson: yep!

PR #201, opened 2022-01-05
Fix casing for page view event name

jowlee:
### Description
Rename the page view event type to be consistent with web


sliptype (approved):
Looks good!

I'm planning to share the event names between client and mobile-client soon so we can avoid future mismatches

PR #202, opened 2022-01-05
[AUD-1194] DynamicImage

sliptype:
### Description

This PR adds the `DynamicImage` component

### Dragons

### How Has This Been Tested?

Tested transitioning between images & placeholders on ios simulator

### DIFF
--- a/src/components/dynamic-image/DynamicImage.tsx
+++ b/src/components/dynamic-image/DynamicImage.tsx
@@ -1,0 +1,146 @@
+import React, { memo, useCallback, useEffect, useRef, useState } from 'react'
+
+import transparentPlaceholderImg from 'audius-client/src/common/assets/image/1x1-transparent.png'
+import useInstanceVar from 'audius-client/src/common/hooks/useInstanceVar'
+import {
+  Animated,
+  Image,
+  ImageSourcePropType,
+  ImageStyle,
+  StyleSheet,
+  View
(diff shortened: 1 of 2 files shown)


src/components/dynamic-image/DynamicImage.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <ImageWithPlaceholder
+          image={firstImage}
+          style={style}
+          usePlaceholder={usePlaceholder}
+        />
+      </Animated.View>
+      <Animated.View
+        style={[
+          styles.image,
+          { opacity: secondOpacity, zIndex: isFirstImageActive ? -1 : 0 }
```
raymondjacobson, line 133: the zIndex here might not work as intended on android. Not 100% sure what to do though b/c elevation adds border shadow, etc. We can def merge this first and play w./ android later if it has issues

raymondjacobson (approved):

src/components/dynamic-image/DynamicImage.tsx
sliptype: Good call, let me test this on android. I initially tried swapping the order of the components in the markup but that was causing some jankiness

sliptype: It does seem to be working on android, going to merge for now and see

PR #203, opened 2022-01-05
Kj add env var for react native navigation

Kyle-Shanks:
### Description

Adding env var for react native nagivation for mobile test flight

### How Has This Been Tested?

Manually tested. Respects env var


raymondjacobson (approved):
mind undoing the package-lock changes since the package.json didn't change in this PR? (we should probably npm install and push a separate commit w/ the update to the lcokfile if it is out of sync)

sliptype (approved):
Looks good!

Since we are not specifying anything in `nativeScreens` and because the RN bottom bar is only showing on native screens right now, we won't see any new features with this flag set to true. 

We could add something to `nativeScreens` if this flag is true, but for example, the feed screen is super early stages right now:
image

What do you guys think?

PR #205, opened 2022-01-06
[AUD-1212] Upgrade react-native types

sliptype:
### Description

This PR upgrades `@types/react-native` to `0.66.3`. I should have done this as part of the react-native upgrade but missed it

This allows us to use `Pressable` and other v66 features

### Dragons

Removes the `display: none` on the AirplayViewManager because the style prop does not exist on it. I tested to make sure it does not display

### How Has This Been Tested?

Checked that there are no type errors. Tested updated components on ios simulator

### DIFF
--- a/src/components/search/TopBar.tsx
+++ b/src/components/search/TopBar.tsx
@@ -183,7 +183,7 @@ const TopBar = ({ onClose, isOpen }: TopBarProps) => {
           onChangeText={onChangeText}
           underlineColorAndroid='transparent'
           style={inputStyles}
-          autoComplete={'off'}
+          autoComplete='off'
           autoCorrect={false}
           returnKeyType={'search'}
           onSubmitEditing={onSubmit}
--- a/src/components/search/TopBar.tsx
+++ b/src/components/search/TopBar.tsx
@@ -183,7 +183,7 @@ const TopBar = ({ onClose, isOpen }: TopBarProps) => {
           onChangeText={onChangeText}
           underlineColorAndroid='transparent'
           style={inputStyles}
-          autoCompleteType={'off'}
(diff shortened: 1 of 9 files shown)


src/components/toast/ToastView.tsx
```diff
@@ -68,8 +68,16 @@ const ToastView = ({ content, timeout, type = 'info' }: ToastViewProps) => {
...
-    Animated.spring(translationAnim, { ...springConfig, toValue: 0 }).start()
+    Animated.spring(opacityAnim, {
+      ...springConfig,
+      toValue: 0,
+      useNativeDriver: true
+    }).start()
+    Animated.spring(translationAnim, {
+      ...springConfig,
+      toValue: 0,
+      useNativeDriver: true
```
dylanjeffers, line 79: That's cool! Is this something we should do for all animations?

src/components/search/TopBar.tsx
```diff
@@ -183,7 +183,7 @@ const TopBar = ({ onClose, isOpen }: TopBarProps) => {
           onChangeText={onChangeText}
           underlineColorAndroid='transparent'
           style={inputStyles}
-          autoCompleteType={'off'}
+          autoComplete={'off'}
```
dylanjeffers, line 186: can maybe add a linter check for this, but ideally `autoComplete='off'`?

dylanjeffers (approved):
Nice changes!

src/components/toast/ToastView.tsx
sliptype: No, it's not supported for certain things like layout animations and gestures:

https://reactnative.dev/docs/animations#caveats

PR #208, opened 2022-01-10
[AUD-1225]: Add ShareDrawer

dylanjeffers:
### Description

Adds native share drawer

### Dragons

It seems like the bump to audius-client resulted in quite a few node_module changes. To ensure I wasn't messing things up, I reset lock-file, removed node_modules, and reinstalled with node version v14.18.2. It still resulted in those amount of changes!

### How Has This Been Tested?

Each action tested on simulator


src/App.tsx
```diff
@@ -37,6 +37,7 @@ import createStore from 'app/store'
 import { setup as setupAnalytics } from 'app/utils/analytics'
 
 import ErrorBoundary from './ErrorBoundary'
+import { ShareDrawer } from './components/share-drawer'
```
sliptype, line 40: Looks like we are using both absolute and relative import paths in this file, let's prefer absolute here?

src/components/share-drawer/ShareDrawer.tsx
```diff
@@ -0,0 +1,199 @@
...
+  modalTitle: 'Share Track',
+  twitter: 'Share to Twitter',
+  tikTok: 'Share Sound to TikTok',
+  copyLink: 'Copy Track to Link'
+}
+
+const createStyles = (themeColors: ThemeColors) =>
+  StyleSheet.create({
+    shareToTwitterAction: {
+      color: '#1BA1F1'
```
sliptype, line 50: It might be worth adding this color to our themes

src/components/share-drawer/utils.ts
```diff
@@ -0,0 +1,6 @@
+export const getTwitterLink = (url: string | null, text: string) => {
```
sliptype, line 1: Can you update `TwitterShare.tsx` to use this function

sliptype (approved):
Awesome! Really liking the `stylesProp`

src/components/share-drawer/ShareDrawer.tsx
dylanjeffers: yes agreed, that was on a list to bring up! so would this go in the `defaultTheme` object?

src/components/share-drawer/utils.ts
dylanjeffers: yeah definitely, regarding sharing utilities like that, i wasnt sure where the best level for this would be, what do you think?

src/components/share-drawer/ShareDrawer.tsx
sliptype: Yeah, you will need to put it in all 3 themes. It could be nice to have some static colors outside of themes though, such as `staticWhite`. But I kinda like having each theme fully declarative

sliptype: Also I want to move the theme definitions out of the utils folder at some point...

PR #212, opened 2022-01-11
[AUD-1161] Track Tile

sliptype:
### Description

This PR:
* Adds the TrackTile and necessary child components
Simulator Screen Shot - iPhone 13 - 2022-01-11 at 16 07 16
* Adds/updates icons for social actions including repost & favorite
* Adds some style utilities in `app/styles` such as `flexRowCentered()`

### Dragons

I covered most edge cases of the track tile but it still may not be 100% there. This will be a good start to building out the lineup though

### How Has This Been Tested?

Tested on multiple sizes of ios simulator, light/dark mode, and android device


src/components/co-sign/CoSign.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+type CoSignProps = {
+  size: Size
+  children: ReactNode
+  style?: StyleProp<ViewStyle>
+}
+
+const CoSign = ({ size, children, style }: CoSignProps) => {
+  const { primary, staticWhite } = useThemeColors()
+  const styles = StyleSheet.create({
```
dylanjeffers, line 18: I might be wrong, but it's not kosher to do stylesheet.create inside render function? my assumption is that you can pass arguments to the Stylesheet.create in the same way we pass the theme in?

src/components/favorite-button/FavoriteButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const FavoriteButton = (props: FavoriteButtonProps) => {
+  const themeVariant = useThemeVariant()
+  const isDarkMode = themeVariant === Theme.DARK
+
+  return (
+    <AnimatedButtonProvider
+      {...props}
+      isDarkMode={isDarkMode}
+      iconLightJSON={[IconFavoriteOnLight, IconFavoriteOffLight]}
```
dylanjeffers, line 34: as an api, it might be nice to have an explicit on/off key, instead of positional args?

src/components/co-sign/CoSign.tsx
sliptype: Oh yeah you're right, I think it would make sense to apply the styles inline in this case. Similar to how it's done for the icon

src/components/favorite-button/FavoriteButton.tsx
sliptype: Icons can have an arbitrary number of states! The repeat button on NowPlaying is one example

src/components/text/Text.tsx
```diff
@@ -27,12 +17,37 @@ const Text = ({ children, weight = 'regular', style, ...props }: Props) => {
...
     >
       {children}
     </RNText>
   )
 }
 
+type AnimatedProps = {
+  children: React.ReactNode
+  weight?: keyof typeof fontByWeight
+} & Parameters<typeof Animated.Text>[0]
```
dylanjeffers, line 31: just curious if there is a better way to get the types for animated here

src/components/track-tile/TrackTile.tsx
```diff
@@ -0,0 +1,190 @@
...
+  return (
+    <View style={styles.container}>
+      {showArtistPick && isArtistPick && (
+        <TrackBannerIcon type={TrackBannerIconType.STAR} />
+      )}
+      {isUnlisted && <TrackBannerIcon type={TrackBannerIconType.HIDDEN} />}
+      <Pressable
+        style={styles.mainContent}
+        onPress={() => {
+          if (showSkeleton) return
```
dylanjeffers, line 129: is there a way to use `disabled` prop for pressable, so we don't have to complicate the onPress function?

src/components/track-tile/TrackTileArt.tsx
```diff
@@ -0,0 +1,74 @@
...
+      userId={coSign.user.user_id}
+    >
+      {imageElement}
+    </CoSign>
+  ) : (
+    <View style={[style, styles.image]}>{imageElement}</View>
+  )
+}
+
+export default memo(TrackTileArt)
```
dylanjeffers, line 74: something ive been seeing a bit, where we default export a memoized comp. from what i can tell we should only use `memo` for places that can truly benefit from perf optimization, and using it can lead to bugs, so just curious of your thoughts there! the other thing that personally makes dev process harder, is the default export results in a component with an undefined name, so in the component heirarchy i see lots of layers of anonymous component names.

src/components/track-tile/TrackTileBottomButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+  onShare,
+  isOwner,
+  isUnlisted,
+  isShareHidden
+}: Props) => {
+  const { toast } = useContext(ToastContext)
+  const { neutralLight4 } = useThemeColors()
+  const styles = useThemedStyles(createStyles)
+
+  const repostButton = () => {
```
dylanjeffers, line 73: these could potentially just be const, what do you think?

src/components/track-tile/TrackTileBottomButtons.tsx
```diff
@@ -0,0 +1,145 @@
...
+    return (
+      <RepostButton
+        onPress={() => toggleRepost()}
```
dylanjeffers, line 76: maybe `onPress={toggleRepost}` to avoid a recompute for the callbacks?

src/components/track-tile/TrackTileCoSign.tsx
```diff
@@ -0,0 +1,83 @@
...
+const formatCoSign = ({
+  hasReposted,
+  hasFavorited
+}: {
+  hasReposted: boolean
+  hasFavorited: boolean
+}) => {
+  if (hasReposted && hasFavorited) {
+    return messages.repostedAndFavorited
+  } else if (hasFavorited) {
```
dylanjeffers, line 30: something that a linter could automate, but technically unecessary else if if we are returning in each block

src/components/track-tile/TrackTileMetadata.tsx
```diff
@@ -0,0 +1,193 @@
...
+import { useThemedStyles } from 'app/hooks/useThemedStyles'
+import { flexCol, flexRow, flexRowCentered } from 'app/styles'
+import { ThemeColors, useThemeColors } from 'app/utils/theme'
+
+import TrackTileArt from './TrackTileArt'
+
+const createStyles = (themeColors: ThemeColors) =>
+  StyleSheet.create({
+    metadata: {
+      ...flexRow()
```
dylanjeffers, line 28: these are fun! do they need to be functions even?

src/components/track-tile/TrackTileMetadata.tsx
```diff
@@ -0,0 +1,193 @@
...
+const messages = {
+  coSign: 'Co-Sign'
+}
+
+type Props = {
+  artistName: string
+  coSign?: Remix | null
+  coverArtSizes: CoverArtSizes
+  fadeIn: { opacity: Animated.Value }
+  goToArtistPage: (e: GestureResponderEvent) => void
```
dylanjeffers, line 100: sometimes i like using the `React.MouseEventHandler` as the full type here, i wonder if there is an equivalent `GestureResponderHandler`?

src/components/track-tile/TrackTileRankIcon.tsx
```diff
@@ -0,0 +1,45 @@
...
+  StyleSheet.create({
+    text: {
+      color: themeColors.secondary
+    },
+    icon: {
+      marginRight: 4
+    }
+  })
+
+const TrackTileRankIcon = ({ showCrown, index, isVisible = true }: Props) => {
```
dylanjeffers, line 32: just wondering about passing in a boolean that indicates `return null` instead of controlling that at the parent level?

src/components/track-tile/TrackTileStats.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const styles = useThemedStyles(createStyles)
+  const trackTileStyles = useThemedStyles(createTrackTileStyles)
+  const { neutralLight4 } = useThemeColors()
+
+  return (
+    <Animated.View style={[fadeIn, styles.stats]}>
+      {isTrending && (
+        <TrackTileRankIcon showCrown={showRankIcon} index={index} />
+      )}
+      {!!(repostCount || saveCount) && !isUnlisted && (
```
dylanjeffers, line 102: nbd but a lil hard to know exactly the conditions here, maybe something like hasEngagement or something?

dylanjeffers (approved):
amazing stuff! this all looks really solid

src/components/text/Text.tsx
sliptype: I don't think there's a clean way to get the props of an Animated component. This also works apparently:

```
Animated.AnimatedProps<React.ComponentPropsWithRef<typeof RNText>>
```

src/components/track-tile/TrackTile.tsx
sliptype: There is indeed, will update!

src/components/track-tile/TrackTileArt.tsx
sliptype: I agree that the `memo` is overkill in most places, I copied this one over from the existing TrackTileArt in the web client. @raymondjacobson would probably be the person to discuss advantages/disadvantages of using memoized components with!

I'm 100% on board with banning default exports, I already have a branch adding that eslint rule. I think the rest of the team is on board too but we can discuss again. Maybe we start doing only named exports moving forward and then we do a sweep of the mobile app and replace all existing default exports

src/components/track-tile/TrackTileBottomButtons.tsx
sliptype: Agreed!

src/components/track-tile/TrackTileMetadata.tsx
sliptype: They don't need to be functions, some of them might take arguments so I figured it would make them easier to use if they were all consistent. `fontWeight` is one example

src/components/co-sign/CoSign.tsx
sliptype: I thought about this more and I'm pretty sure it is kosher to do stylesheet.create in the render function. We're already doing that (memoized) every time we use `useThemedStyles`

The main benefit to using stylesheet.create  is to get type checking of styles before you apply them to a component, I don't think there's anything expensive happening in there

I'm still gonna update this though because it will be cleaner without it

src/components/co-sign/CoSign.tsx
```diff
@@ -0,0 +1,90 @@
...
+  style?: StyleProp<ViewStyle>
+}
+
+const styles = StyleSheet.create({
+  check: {
+    position: 'absolute'
+  }
+})
+
+const layoutBySize = {
```
Kyle-Shanks, line 22: I like this much better 👌

src/styles/typography.ts
```diff
@@ -0,0 +1,16 @@
...
+  bold: 'AvenirNextLTPro-Bold',
+  demiBold: 'AvenirNextLTPro-DemiBold',
+  medium: 'AvenirNextLTPro-Medium',
+  regular: 'AvenirNextLTPro-Regular',
+  light: 'AvenirNextLTPro-Light',
+  thin: 'AvenirNextLTPro-Thin',
+  ultraLight: 'AvenirNextLTPro-UltLt'
+}
+
+export const fontWeight = (weight: keyof typeof fontByWeight): TextStyle => ({
```
Kyle-Shanks, line 14: Thinking a name like `fontFromWeight` or something would be a bit more clear here

src/styles/flex.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const flexRow = (): ViewStyle => ({
+  display: 'flex',
+  flexDirection: 'row'
+})
+
+export const flexRowCentered = (): ViewStyle => ({
+  ...flexRow(),
+  alignItems: 'center'
+})
```
Kyle-Shanks, line 20: 🔥

src/styles/typography.ts
sliptype: What about just `font`? I feel like we could add things like letter spacing to this too

Kyle-Shanks: Yea `font` or `fontStyle` or something like that works for me. I felt that fontWeight was a bit less clear

src/components/track-tile/TrackTileMetadata.tsx
dylanjeffers: ahh perfect, thanks

Kyle-Shanks (approved):
Nice

PR #216, opened 2022-01-12
Add optional icon to drawer title

rickyrombo:
### Description

Adds an (optional) `titleIcon` prop to the `Drawer` component to allow drawers to have icons in the header. Requires title to be present.

### Dragons

Should the `titleIcon` prop take an element or callback instead of the `ImageSourcePropType`? Thinking about SVGs here...

### How Has This Been Tested?

Locally tested using upcoming `ChallengeRewardsDrawer` changes, but with both icon included and not included.

N/A

### DIFF
--- a/src/components/drawer/Drawer.tsx
+++ b/src/components/drawer/Drawer.tsx
@@ -10,6 +10,9 @@ import {
   Animated,
   Dimensions,
   GestureResponderEvent,
+  Image,
+  ImageSourcePropType,
+  ImageStyle,
   LayoutChangeEvent,
   PanResponder,
   PanResponderGestureState,
@@ -65,7 +68,7 @@ const createStyles = (zIndex = 5) => (themeColors: ThemeColors) =>
       height: '100%'
     },
 
-    titleContainer: {
+    titleBarContainer: {
       display: 'flex',
       flexDirection: 'row',
       alignItems: 'center',
@@ -79,10 +82,23 @@ const createStyles = (zIndex = 5) => (themeColors: ThemeColors) =>
       left: 24
     },
 
+    titleContainer: {
+      display: 'flex',
+      flexDirection: 'row',
+      alignItems: 'center',
+      justifyContent: 'center',
+      paddingTop: 20
+    },
+
+    titleIcon: {
+      marginRight: 12,
+      height: 24,
+      width: 24
+    },
+
     titleLabel: {
       fontFamily: 'AvenirNextLTPro-Bold',
       fontSize: 18,
-      paddingTop: 20,
       color: themeColors.neutral
     },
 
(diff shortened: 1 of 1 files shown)


dylanjeffers (commented):
this is great, i kinda had to implement a version of this using the `renderTitle` prop for ActionDrawer, would this make that much easier? if so would be great to refactor that case, and at least also pass through `titleIcon` for ActionDrawer. for ref: https://github.com/AudiusProject/audius-mobile-client/blob/b2c6821ab5f014db03742b8c271d720f4c9c9b56/src/components/share-drawer/ShareDrawer.tsx#L200

src/components/drawer/Drawer.tsx
```diff
@@ -233,19 +253,21 @@ const attachToDy = (animation: Animated.Value, newValue: number) => (
 const DrawerHeader = ({
   onClose,
   title,
+  titleIcon,
   isFullscreen,
   zIndex
 }: {
   onClose: () => void
   title?: string
+  titleIcon?: ImageSourcePropType
```
dylanjeffers, line 262: there is a fun opportunity to strongly type the requirement that titleIcon can only be present when title is: https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#props-pass-one-only-if-the-other-is-passed

src/components/drawer/Drawer.tsx
```diff
@@ -135,6 +151,10 @@ export type DrawerProps = {
    * Header title if this drawer has a header
    */
   title?: string
+  /**
+   * Icon to display in the header next to the title (must also include title)
+   */
+  titleIcon?: ImageSourcePropType
```
dylanjeffers, line 157: there is a fun opportunity to strongly type the requirement that titleIcon can only be present when title is: https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#props-pass-one-only-if-the-other-is-passed

rickyrombo:
> this is great, i kinda had to implement a version of this using the `renderTitle` prop for ActionDrawer, would this make that much easier? if so would be great to refactor that case

Aha, I was wondering where I did this last time! I knew I'd seen something like this before and went the renderProp route. I don't think `ActionDrawer` even uses the `Drawer` `title` prop, though. This might take some thought/care

dylanjeffers:
> > this is great, i kinda had to implement a version of this using the `renderTitle` prop for ActionDrawer, would this make that much easier? if so would be great to refactor that case
> 
> Aha, I was wondering where I did this last time! I knew I'd seen something like this before and went the renderProp route. I don't think `ActionDrawer` even uses the `Drawer` `title` prop, though. This might take some thought/care

ah thats so right, it doesnt even use the title prop, oh my :D, def something to consider at a later date

dylanjeffers (approved):
LGTM, optional typing idea for the icon

src/components/drawer/Drawer.tsx
rickyrombo: I like it!

PR #217, opened 2022-01-13
Improve webApp render flow and missing store edge-cases

dylanjeffers:
### Description

- Adds WebAppManager to establish requirement that the web-app needs to render above native app
- Adds a check in render flow to ensure clientStore is available before rendering the rest of the native app. This simplifies `useSelectorWeb` implementation, and prevents bugs where data returned by useSelectorWeb is undefined for the first render. Weirdly enough typescript wasn't picking up the fact that `useSelectorWeb` can return undefined, so something like `const {a, b} = useSelectorWeb(someSelector)` results in a runtime error, `can't resolve a of undefined`

### Dragons

This change affects the root of the app, so there is a potential for things to break in weird ways.

### DIFF
--- a/src/hooks/useSelectorWeb.ts
+++ b/src/hooks/useSelectorWeb.ts
@@ -1,5 +1,4 @@
 import { CommonState } from 'audius-client/src/common/store'
-import { isEmpty } from 'lodash'
 import { Selector, useSelector } from 'react-redux'
 
 // When mobile client is no longer dependent on the web client
@@ -8,6 +7,6 @@ export const useSelectorWeb = <ReturnValue>(
   selector: Selector<CommonState, ReturnValue>
 ) => {
   return useSelector((state: { clientStore: CommonState }) =>
-    isEmpty(state.clientStore) ? undefined : selector(state.clientStore)
+    selector(state.clientStore)
   )
 }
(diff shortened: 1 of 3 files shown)


src/hooks/useSelectorWeb.ts
```diff
@@ -8,6 +7,6 @@ export const useSelectorWeb = <ReturnValue>(
   selector: Selector<CommonState, ReturnValue>
 ) => {
   return useSelector((state: { clientStore: CommonState }) =>
-    isEmpty(state.clientStore) ? undefined : selector(state.clientStore)
+    selector(state.clientStore)
```
raymondjacobson, line 10: Not sure this is a better option -- but maybe worth calling out, we *could* return `{}` instead of delaying render. We'd have to check downstream if the derived fields (a & b in the case of { a, b } = useSelectorWeb()) aren't undefined, but it might reduce UI headache if there is any on initial paint

raymondjacobson (approved):

src/hooks/useSelectorWeb.ts
dylanjeffers: yes we actually tried this! but that actually breaks other components that arent expectind an object, like
```
const someString = useSelectorWeb(getThatString)
```

raymondjacobson: of course. carry on

PR #218, opened 2022-01-13
[AUD-1273] consume new share state, handle collections

dylanjeffers:
### Description

Uses the new share-modal state, which enables sharing artists, tracks, and collections.

### DIFF
--- a/src/components/share-drawer/ShareDrawer.tsx
+++ b/src/components/share-drawer/ShareDrawer.tsx
@@ -32,15 +32,8 @@ import {
 import ActionDrawer from '../action-drawer'
 import { ToastContext } from '../toast/ToastContext'
 
+import { messages } from './messages'
 import { getContentUrl, getTwitterShareUrl } from './utils'
-
-const messages = {
-  modalTitle: 'Share Track',
-  twitter: 'Share to Twitter',
-  tikTok: 'Share Sound to TikTok',
-  copyLink: 'Copy Track to Link',
-  shareSheet: 'Share Track via...'
-}
 
 const createStyles = (themeColors: ThemeColors) =>
   StyleSheet.create({
@@ -96,6 +89,7 @@ export const ShareDrawer = () => {
       account &&
       account.user_id === content.artist.user_id
   )
+  const shareType = content?.type ?? 'track'
 
   const handleClose = useCallback(() => {
     dispatchWeb(setVisibility({ modal: 'Share', visible: false }))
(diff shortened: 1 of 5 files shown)


src/components/share-drawer/ShareDrawer.tsx
```diff
@@ -89,55 +79,76 @@ export const ShareDrawer = () => {
...
+        break
+    }
     handleClose()
-  }, [dispatchWeb, track, source, handleClose])
-
-  const shouldIncludeTikTokAction =
-    isOwner && !track.is_invalid && !track.is_delete
+  }, [dispatchWeb, content, source, handleClose])
+
+  const shouldIncludeTikTokAction = Boolean(
```
rickyrombo, line 145: What does `Boolean()` do here - isn't it already a boolean?

src/components/share-drawer/messages.ts
```diff
@@ -0,0 +1,25 @@
+import { ShareType } from 'audius-client/src/common/store/ui/share-modal/types'
+
+const shareTypeMap: Record<ShareType, string> = {
+  track: 'Track',
+  profile: 'Profile',
+  album: 'Album',
+  playlist: 'Playlist'
+}
+
+export const messages = {
```
rickyrombo, line 10: Just thinking ahead to when we get around to actually using something like [l18n](https://react.i18next.com/), we'll probably need these to not be functions but plain strings which are then interpolated after the fact. I'm not sure we need it to work that way right now, since I don't know which internationalization tool we're going to use and when, but worth considering when making these `messages` objects.

Maybe we roll our own interpolator in the meantime? Might be something to add to prod eng sync

src/components/share-drawer/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+import {
+  getCollectionRoute,
+  getTrackRoute,
+  getUserRoute
+} from 'app/utils/routes'
+import { getTwitterLink } from 'app/utils/twitter'
+
+import { messages } from './messages'
+
+export const getContentUrl = content => {
```
rickyrombo, line 12: Should this be typed?

src/components/share-drawer/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+      return getCollectionRoute({ ...album, user: artist }, true)
+    }
+    case 'playlist': {
+      const { playlist, creator } = content
+      return getCollectionRoute({ ...playlist, user: creator }, true)
+    }
+  }
+}
+
+export const getTwitterShareText = content => {
```
rickyrombo, line 33: Same here - just so we know we need a `ShareModalContent`?

rickyrombo (approved):
Looks great! Minor nits in my review comments

src/components/share-drawer/utils.ts
dylanjeffers: Yeah absolutely, we should def get eslint complaining

src/components/share-drawer/messages.ts
dylanjeffers: yeah totally agreed, would be open to hear what folks think at a prod sync

src/components/share-drawer/ShareDrawer.tsx
dylanjeffers: yeah no reason here! in audius-client i was seeing it was like `boolean | null` which was messing up the prop type of just `boolean`

PR #219, opened 2022-01-13
Upload Bounce on master

sliptype:
### Description

This PR:
* Updates CI to upload the Bounce app to testflight on pushes to `master` instead of release branches

This means that Bounce will always have latest master and we don't need to create a release branch to test features. We could potentially include release branches in the regex too, the only use case I can see for that is a hotfix on a release branch that we want to test on Bounce. Not sure if this would ever happen

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,8 +256,7 @@ jobs:
           env: '.env.stage'
       - when:
           condition:
-            matches:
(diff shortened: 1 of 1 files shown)


raymondjacobson (approved):

dylanjeffers:
awesome!

PR #222, opened 2022-01-14
[AUD-1278] Use common models

dylanjeffers:
### Description

Replaces the duplicate models with models from audius-client common

### Dragons

It is a bit weird that we have special subsets of models still in the `models` directory, (for example https://github.com/AudiusProject/audius-mobile-client/pull/222/files#diff-1f530071e85873610cec5e769763b0e80030298e9d8a6eda72037e5507698e63) would be good to figure out what to do with them. This pr also updates the paths to some of the audius-client dependencies that are not in `common` ie changing `components` to `pages`, that is a result of the audius-client directory restructuring


sliptype (approved):
Nice!

Yeah I think we could probably moves those other types into common in the future. Because they aren't really specific to mobile, they only happen to be used in mobile

PR #223, opened 2022-01-14
[AUD-1278] Use common analytics types

dylanjeffers:
### Description

Removes duplicate analytics types and uses the common ones from `audius-client`

### Dragons

There is a bit of funkiness required to merge the common enums with the native specific ones. One one hand it makes sense to have native specific extensions, but i can also see the argument to put them all in common. I'm up for either so lmk. Mobile also uses different naming conventions for analytics, like `Name.AnalyticsEvent` vs `EventNames.AnalyticsEvent`, and `AllEvents` vs `AllAnalyticsEvents`. happy to approach unifying those names now if it makes sense

### DIFF
--- a/src/types/analytics.ts
+++ b/src/types/analytics.ts
@@ -1,6 +1,29 @@
-import { Message } from '../message/types'
+import {
+  Name as CommonEventNames,
+  AllTrackingEvents as CommonTrackingEvents
+} from 'audius-client/src/common/models/Analytics'
 
-type JsonMap = Record<string, unknown>
+import { Message } from 'app/message'
+
+export * from 'audius-client/src/common/models/Analytics'
+
+enum MobileEventNames {
+  NOTIFICATIONS_OPEN_PUSH_NOTIFICATION = 'Notifications: Open Push Notification'
+}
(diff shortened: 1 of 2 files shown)


src/types/analytics.ts
```diff
@@ -1,6 +1,29 @@
-import { Message } from '../message/types'
+import {
+  Name as CommonEventNames,
+  AllTrackingEvents as CommonTrackingEvents
+} from 'audius-client/src/common/models/Analytics'
 
-type JsonMap = Record<string, unknown>
+import { Message } from 'app/message'
+
+export * from 'audius-client/src/common/models/Analytics'
```
sliptype, line 8: So this means we are export both `Name` and `EventNames` from here, that might be a bit confusing. Probably best not to re-export `Name`

sliptype (approved):
Nice work! Makes sense to keep the native specific stuff here IMO

src/types/analytics.ts
dylanjeffers: yeah good call, will explicitely export the needed items

PR #229, opened 2022-01-18
[AUD-1303] Fix warnings & deprecations

sliptype:
### Description

This PR:
* Fixes usage of deprecated `removeEventListener`
* Fixes deprecated forking of sagas
* Fixes warning caused by calling `useDarkMode()` too often
* Changes web app errors to log as warnings, to avoid triggering fullscreen LogBox
* Fixes react devtools in react-native-debugger by installing correct version of `react-devtools-core`

### Dragons

Web App errors are now being logged as warnings instead of errors! This should only have an effect when developing, as web app errors didn't crash the app to begin with

### How Has This Been Tested?

iOS simulator


dylanjeffers (approved):
This is great! No more annoying errors, and the debugger works now

PR #231, opened 2022-01-19
[AUD-1313] Null check notification entity when getting route

sliptype:
### Description

This PR adds some null checks when getting the route for a notification. This is related to this error: https://sentry.io/organizations/audius/issues/2871942483/?project=1457231&query=sdk.name%3Asentry.javascript.react-native&sort=freq&statsPeriod=14d

### Dragons

### How Has This Been Tested?

### DIFF
--- a/src/components/notifications/content/Entity.tsx
+++ b/src/components/notifications/content/Entity.tsx
@@ -24,7 +24,10 @@ type EntityProps = {
 
 const Entity = ({ entity, entityType, onGoToRoute }: EntityProps) => {
   const onPress = useCallback(() => {
-    onGoToRoute(getEntityRoute(entity, entityType))
+    const route = getEntityRoute(entity, entityType)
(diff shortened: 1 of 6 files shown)


rickyrombo (approved):
Curious, why are `null` tracks being put in the `getEntityRoute` function in the first place? Is this going to mask "broken" notifications or is it really expected these could be null?

sliptype:
> Curious, why are `null` tracks being put in the `getEntityRoute` function in the first place? Is this going to mask "broken" notifications or is it really expected these could be null?

I thought this was something we were seeing on the client too but now I think that's not the case. I'm digging further into it,  could be something to do with how the notification is connected from the cache before being passed to native. Gonna see if I can find the root cause

sliptype:
Closing this for now. Creating another PR that adds logging

PR #213, opened 2022-01-12
[AUD-1223] [AUD-1005] Add ChallengeRewardsDrawer and CognitoDrawer

rickyrombo:
### Description

~**NOTE:** This is based on #206~ Updated this PR to be based on master

AUD-1005 AUD-1223

- Adds the `ChallengeRewardsDrawer` that mimics the `ChallengeRewardsModal` on web
- Adds a new `CognitoDrawer` modal component that uses a webview (note: similar to `CognitoModal` on web, but different! see below)

Note: Follows a decent amount of changes on `audius-client` to make this smoother
- [x] **TODO:** Update client version once https://github.com/AudiusProject/audius-client/pull/794 goes in

Screenshots:
image
image

### Dragons

*Is there anything the reviewer should be on the lookout for? Are there any dangerous changes?*

- This is a bit different from the existing mobile Cognito flow, which uses a separate `window.Flow` API. Instead, this one calls Identity to get an embeddable/shareable URL for use with a mobile webview. See: https://github.com/AudiusProject/audius-protocol/pull/2284
- Note that the HCaptcha does not require the challenge modal to be hidden on mobile, but I left that the same for consistency/code-reuse.
- Progress bar has _slightly_ different look, since an inset shadow isn't available in React Native. Imitated best I could.

### How Has This Been Tested?

*Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration.*

- Tested on iOS simulator manually using a latest master webclient and staging backend.

### How will this change be monitored?

*For features that are critical or could fail silently please describe the monitoring/alerting being added.*

N/A - Errors in libs for claiming rewards should get passed through to sentry

PR #234, opened 2022-01-19
Use hook for press scale animation on buttons

rickyrombo:
### Description

Adds a new generic hook `usePressScaleAnimation` to replace existing animation sequences for buttons

### Dragons

### How Has This Been Tested?

Tested a settings button and my copy button from #213 on iOS simulator

N/A

### DIFF
--- a/src/components/audio-breakdown-drawer/AudioBreakdownDrawer.tsx
+++ b/src/components/audio-breakdown-drawer/AudioBreakdownDrawer.tsx
@@ -317,7 +317,7 @@ const Wallet = ({ chain, address, balance }: WalletProps) => {
   const solWalletAudioEnabled = false
   const styles = useThemedStyles(createStyles)
 
-  const { scale, startPress, stopPress } = usePressScaleAnimation(0.98)
+  const { scale, handlePressIn, handlePressOut } = usePressScaleAnimation(0.98)
 
   const displayAddress =
     chain === Chain.Eth ? shortenEthAddress : shortenSPLAddress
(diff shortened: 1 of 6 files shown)


src/hooks/usePressScaleAnimation.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { Animated } from 'react-native'
+
+export const usePressScaleAnimation = (scaleTo = 0.97) => {
+  const scale = useRef(new Animated.Value(1)).current
+
+  const startPress = useCallback(() => {
+    Animated.timing(scale, {
+      toValue: scaleTo,
+      duration: 100,
```
raymondjacobson, line 11: Can we switch this to spring physics or use easing https://reactnative.dev/docs/easing#ease

sliptype (approved):
Cleeeean

src/hooks/usePressScaleAnimation.ts
raymondjacobson: and for web consistency, 
https://github.com/AudiusProject/stems/blob/master/src/components/Button/Button.module.css#L8
https://github.com/AudiusProject/stems/blob/master/src/assets/styles/animations.css#L9

src/components/audio-breakdown-drawer/AudioBreakdownDrawer.tsx
```diff
@@ -316,7 +317,7 @@ const Wallet = ({ chain, address, balance }: WalletProps) => {
   const solWalletAudioEnabled = false
   const styles = useThemedStyles(createStyles)
 
-  const scale = useRef(new Animated.Value(1)).current
+  const { scale, startPress, stopPress } = usePressScaleAnimation(0.98)
```
raymondjacobson, line 320: might be nice to have these `{ scale, onPressIn, onPressOut }` so it's super obvious where they should go

src/hooks/usePressScaleAnimation.ts
rickyrombo: It's already using easing in-out by default, but if you want I can make it explicit?
https://reactnative.dev/docs/animated#configuring-animations
> Animated.timing() animates a value over time using easing functions.
> By default, it uses a symmetric easeInOut curve that conveys the gradual acceleration of an object to full speed and concludes by gradually decelerating to a stop.

raymondjacobson: nice nvm

raymondjacobson (approved):

PR #237, opened 2022-01-20
[AUD-1329]Add fastlane device refresh script to run when updating the list of devices

milindNshah:
### Description
Add npm script: ios:device-refresh to update list of devices when new ones are added. 
There will still be a lot of manual steps however.

### How Has This Been Tested?
Ran the script, working as expected


sliptype (approved):
Nice!

Maybe since the error message is "Signing certificate is invalid..." we could have the command be something to the effect of `ios:update-signing-cert`

dylanjeffers (approved):
great idea

package.json
```diff
@@ -17,6 +17,7 @@
     "ios:device:bounce": "react-native run-ios --scheme 'Bounce' --configuration Release --device",
     "ios:dev": "react-native run-ios --scheme 'Dev'",
     "ios:device:dev": "react-native run-ios --scheme 'Dev' --device",
+    "ios:device-refresh": "cd ios && fastlane match development -a co.audius.audiusmusic,co.audius.audiusmusic.bounce",
```
raymondjacobson, line 20: nit, should we `cd -` or `cd ..` to go back to the same directory after this runs?

raymondjacobson (approved):

PR #238, opened 2022-01-20
[AUD-1317] Catch "Failed to delete file" error

sliptype:
### Description

This PR addresses this sentry log: https://sentry.io/organizations/audius/issues/2899570741/?query=sdk.name%3Asentry.javascript.react-native+message%3A%22failed+to+delete%22&statsPeriod=14d

In the case of an error, the file is deleted. But sometimes this fails and throws an error. I've wrapped this in an empty try catch because if the file fails to delete, it's probably a filesystem issue (looks like it mostly happens on android) and is pretty inconsequential for the app

Also took the opportunity to split the download logic into a separate function for ease of reuse!

In the process I noticed that files on android where being saved with `download successful!` appended to them. (Thought this was maybe what was causing the file delete failure error, but that didn't seem to be the case.) Updating the notification title in the RNFetchBlob config fixed this. And the notification was still informative enough:

image

image

### Dragons

### How Has This Been Tested?

Tested on ios and android devices


src/message/handlers/download.ts
```diff
@@ -19,6 +23,52 @@ const cancelDownloadTask = () => {
...
+    fetchTask.progress({ interval: 250 }, (received, total) => {
+      dispatch(setDownloadedPercentage((received / total) * 100))
+    })
+
+    const fetchRes = await fetchTask
+
+    // Do this after download is done
+    dispatch(setVisibility({ drawer: 'DownloadTrackProgress', visible: false }))
+
+    await onFetchComplete?.(fetchRes)
```
milindNshah, line 61: This is sick! I didn't know you could use nullable functions like that. Just want to confirm that this actually does work on iOS (i.e, passes the parameter correctly?)

milindNshah (approved):
Dope, thanks for fixing and refactoring!

src/message/handlers/download.ts
sliptype: The optional chaining operator definitely looks a little wonky on functions lol, but it's super useful!

And yup I just double checked that the param gets passed properly on ios and sharing + flushing is successful

PR #243, opened 2022-01-21
[AUD-1283] Create forgot password drawer

milindNshah:
### Description
- Create new drawer for forgot password helper
- Add new question mark icon
- Add text and some UI changes in signin/signup page

### How Has This Been Tested?
Screen Shot 2022-01-21 at 3 11 46 PM
Screen Shot 2022-01-21 at 3 12 07 PM
Screen Shot 2022-01-21 at 3 12 00 PM
__iphone12-mini__
Screen Shot 2022-01-24 at 10 20 10 AM


src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
```diff
@@ -0,0 +1,155 @@
...
+    fromContent: {
+      color: themeColors.neutral,
+      fontWeight: '600',
+      fontSize: 16,
+      lineHeight: 20,
+      marginBottom: 12
+    },
+
+    subjectContent: {
+      color: themeColors.neutral,
```
dylanjeffers, line 101: general thought here for us, it might be helpful for us to have some established typography, so we can build off those and make sure our designs use consistent values, what do you think? i have a few ideas on how we can represent it

src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
```diff
@@ -0,0 +1,155 @@
...
+    'To restore access to your account, search for the email we sent when you first signed up.',
+  fromHeader: 'From:',
+  subjectHeader: 'Subject:',
+  from: 'recovery@audius.co',
+  subject: '"Save This Email: Audius Password Recovery"'
+}
+
+const ForgotPasswordDrawer = () => {
+  const [isOpen, setIsOpen] = useDrawer('ForgotPassword')
+  console.log('rendering drawer', isOpen)
```
dylanjeffers, line 120: console

src/components/signon/SignOn.tsx
```diff
@@ -512,6 +523,19 @@ const SignOn = ({ navigation }: SignOnProps) => {
     )
   }
 
+  const ForgotPasswordButton = () => {
```
dylanjeffers, line 526: looks like this was already a pattern in this file, but we should avoid creating components in the render body of other components. this actually leads to a big perf hit as react can't cache these components. 

some options we can do are:
- define these outside the signon component
- make these inline functions. to do that, we should use camelCase, and probably change the name to include render, and doing `{renderForgotPasswordButton()}` instead of `<ForgotPasswordButton />`
- we just make this a constant with `const forgotPasswordButton = (<Text ... />)`

dylanjeffers (approved):
looks great! i think we should figure out a plan for those inner defined components, lmk what you think

src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { useThemeColors, ThemeColors } from 'app/utils/theme'
+
+const createStyles = (themeColors: ThemeColors) =>
+  StyleSheet.create({
+    mainView: {
+      display: 'flex',
+      flexDirection: 'column',
+      paddingTop: 24,
+      paddingHorizontal: 24,
+      fontFamily: 'AvenirNextLTPro-Regular'
```
sliptype, line 19: I don't think this fontFamily has any effect since it's not on a Text component

src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
sliptype: @milindNshah You shouldn't have to specify the color as neutral, the Text component does that by default

@dylanjeffers It would be ideal for sure, but this is another one of those cases where our designs aren't necessarily consistent enough to make common patterns valuable. But maybe having the common patterns would push us towards more consistent designs. For now, the `Text` component sets the fontFamily + color

sliptype (approved):
Awesome! One thing I would check is how this looks on smaller screen sizes, you can spin up a simulator for an iphone mini or similar

src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
dylanjeffers: Thanks for the context!

src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
milindNshah: RIP, good catch

src/components/signon/SignOn.tsx
milindNshah: Yeah makes sense, i think the 2nd option is what the general codebase patterns follow so I'll update to that

milindNshah:
> Awesome! One thing I would check is how this looks on smaller screen sizes, you can spin up a simulator for an iphone mini or similar

Added a screenshot with the iphone12 mini!

PR #249, opened 2022-01-26
[AUD-1162] Lineup

sliptype:
### Description

This PR: 
  * Adds the `Lineup` component
    * Infinite loading w/ skeletons
    * Playing/enqueuing tracks (via client actions)
    * Delineation by time
Simulator Screen Shot - iPhone 13 - 2022-01-26 at 16 39 03
  * Adds example Lineup to Feed and Track screens
  * Updates TrackTile to fetch it's own data based on the uid
  * Temporarily uses `isEqual` as an equality function for some `useSelectorWeb` to prevent rerenders. The plan is to improve this: https://linear.app/audius/issue/AUD-1364/reduce-useselector-rerenders
  * Updates shadows of TrackTile to use react-native-shadow-2
  * Adds a `TrackTileSkeleton` component to simplify data flow when showing loading tracks (no need for fallbacks)
  * Removes unnecessary flex utilities
 
The lineup component isn't 100% complete, but close enough to unblock pages. Still need to implement:
  * Self loading variant (Right now the data is always based on client state so it isn't immediately necessary)
  * Featured track
  * Playlist tiles
  * Styling of the pull to refresh indicator
  * AnimatedButton animation upon state changes

Question: Do we still need to account for deleted tracks coming back from the api when paginating?

### Dragons

* We will have to be careful to load the corresponding lineup in the client until we separate the clients. (Might be resolved via the self loading variant)
* `useImageSize` still seems to be breaking when not `yalc`ed. Going to look at this next

### How Has This Been Tested?

Tested on iOS simulator


src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+  const playingUid = useSelector(getPlayingUid)
+  const itemCounts = useItemCounts(variant)
+
+  // Item count based on the current page
+  const pageItemCount = useMemo(() => {
+    return itemCounts.initial + (lineup.page - 1) * itemCounts.loadMore
+  }, [itemCounts, lineup])
+
+  // Either the provided count or a default
+  const countOrDefault = useMemo(
```
dylanjeffers, line 120: just curious if this needs to be a memo? i believe nums can be recomputed without triggering rerender?

src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+  const dispatchWeb = useDispatchWeb()
+
+  const playing = useSelector(getPlaying)
+  const playingUid = useSelector(getPlayingUid)
+  const itemCounts = useItemCounts(variant)
+
+  // Item count based on the current page
+  const pageItemCount = useMemo(() => {
+    return itemCounts.initial + (lineup.page - 1) * itemCounts.loadMore
+  }, [itemCounts, lineup])
```
dylanjeffers, line 117: just curious if the lineup changes at all, or that object is stable? if not, we could just put lineup.page in the deps?

src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+    if (!hasMore) {
+      return
+    }
+
+    const lineupLength = lineup.entries.length
+    const offset = lineupLength + lineup.deleted
+
+    if (
+      (!limit || lineupLength !== limit) &&
```
dylanjeffers, line 138: kinda a lot here, prob fine, but could be a nice constant?

src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+            id: `${id}`,
+            source: source || PlaybackSource.TRACK_TILE
+          })
+        )
+      }
+    },
+    [playTrack, pauseTrack, playing, playingUid]
+  )
+
+  const renderItem = ({ index, item }: { index: number; item: LineupItem }) => {
```
dylanjeffers, line 189: not sure if it makes sense to make this its own component that can dispatch on the item type? could be helpful for reducing complexity of the page. also a potential idea is to pass the render methods into the lineup, and the lineup's core logic is managing pagination etc?

src/components/lineup/delineate.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const Delineations = Object.freeze({
+  TODAY: 'today',
+  YESTERDAY: 'yesterday',
+  EARLIER_THIS_WEEK: 'earlier this week',
+  LAST_WEEK: 'last week',
+  EARLIER_THIS_MONTH: 'earlier this month'
+})
+
+const getGroup = timestamp => {
```
dylanjeffers, line 21: potentially a better name?

src/components/lineup/delineate.ts
```diff
@@ -0,0 +1,56 @@
...
+  return startOfMonth.format('MMMM YYYY')
+}
+
+export const delineateByTime = (entries: LineupItem[]) => {
+  return Object.entries(groupBy<LineupItem>(entries, getGroup)).map(
+    ([title, data], index) => ({
+      // For the first group,
+      // set the title as empty string to prevent
+      // Delineator from being displayed
+      title: index === 0 ? '' : title,
```
dylanjeffers, line 52: could we filter the first one out? probably missing some context here :)

src/components/lineup/types.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export enum LineupVariant {
+  MAIN = 'main',
+  SECTION = 'section',
+  CONDENSED = 'condensed',
+  PLAYLIST = 'playlist'
+}
+
+export type LineupItem =
+  | {
```
dylanjeffers, line 16: personally kinda like making these their own types, but maybe hard to name the loaded state?

src/components/track-tile/TrackTile.tsx
```diff
@@ -1,106 +1,215 @@
...
+  // not being preserved when syncing redux state from client.
+  // This can be removed when no longer dependent on web client
+  const track = useSelectorWeb(state => getTrack(state, { uid }), isEqual)
+  const user = useSelectorWeb(state => getUserFromTrack(state, { uid }))
+  if (!track || !user) {
+    console.warn('Track or user missing for TrackTile, preventing render')
+    return null
+  }
+
+  return <TrackTileComponent {...props} track={track} user={user} />
```
dylanjeffers, line 74: is the main reason to split these up so we can be sure track and user are there before including the other hooks in TrackTileComponent?

src/components/track-tile/TrackTile.tsx
```diff
@@ -1,106 +1,215 @@
...
-  toggleSave,
-  user,
-  uid
-}: TrackTileProps) => {
+  track,
+  uid,
+  user
+}: TrackTileProps & { track: Track; user: User }) => {
+  const {
+    // permalink,
```
dylanjeffers, line 89: comment

src/components/track-tile/TrackTile.tsx
```diff
@@ -1,106 +1,215 @@
...
+    track_id
+  } = track
+  const { _artist_pick, name, user_id } = user
+  const navigation = useNavigation<
+    NativeStackNavigationProp<BaseStackParamList>
+  >()
+
+  const dispatch = useDispatch()
+  const dispatchWeb = useDispatchWeb()
+  // const pushRouteWeb = usePushRouteWeb()
```
dylanjeffers, line 111: same

src/components/track-tile/TrackTile.tsx
```diff
@@ -1,106 +1,215 @@
...
 
-  const styles = useThemedStyles(createStyles)
-  const hideShare: boolean = fieldVisibility
-    ? fieldVisibility.share === false
+  const hideShare: boolean = field_visibility
+    ? field_visibility.share === false
     : false
-  const hidePlays = fieldVisibility
-    ? fieldVisibility.play_count === false
+  const hidePlays = field_visibility
```
dylanjeffers, line 129: would `field_visibility?.play_count === false` work?

src/components/track-tile/TrackTile.tsx
```diff
@@ -109,70 +218,65 @@ export const TrackTile = ({
     }
   }, [onLoad, isLoaded, index, opacity])
 
+  if (is_delete || user?.is_deactivated) {
```
dylanjeffers, line 221: maybe we do these checks in the parent component?

src/components/track-tile/TrackTile.tsx
```diff
@@ -109,70 +218,65 @@ export const TrackTile = ({
...
-          toggleSave={onToggleSave}
-          onShare={onClickShare}
-          onPressOverflow={onClickOverflowMenu}
+          hasReposted={has_current_user_reposted}
+          hasSaved={has_current_user_saved}
           isOwner={isOwner}
-          isUnlisted={isUnlisted}
           isShareHidden={hideShare}
+          isUnlisted={is_unlisted}
+          onPressOverflow={onPressOverflowMenu}
```
dylanjeffers, line 274: would it make sense to define all these callbacks in TrackTlieBottomButtons?

src/components/track-tile/TrackTileContainer.tsx
```diff
@@ -0,0 +1,38 @@
...
+    },
+    mainContent: {
+      flex: 1
+    }
+  })
+
+export const TrackTileContainer = ({ children }: { children: ReactNode }) => {
+  const styles = useThemedStyles(createStyles)
+  return (
+    <Shadow
```
dylanjeffers, line 29: woohoo!

src/components/track-tile/TrackTileMetadata.tsx
```diff
@@ -107,73 +66,54 @@ export const TrackTileMetadata = ({
...
-          <AnimatedText
-            style={[
-              fadeIn,
-              styles.titleText,
-              isPlaying ? styles.titlesActive : {}
-            ]}
+      <View style={trackTileStyles.titles}>
+        <Pressable style={trackTileStyles.title} onPress={goToTrackPage}>
+          <Text
+            style={[styles.titleText, isPlaying ? styles.titlesActive : {}]}
```
dylanjeffers, line 93: i believe you can just do `isPlaying && styles.titlesActive`

src/components/track-tile/TrackTileSkeleton.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react'
+
+import { ImageStyle, StyleSheet, View } from 'react-native'
```
dylanjeffers, line 3: this file is perfect

src/components/track-tile/TrackTileSkeleton.tsx
```diff
@@ -0,0 +1,60 @@
...
+})
+
+export const TrackTileSkeleton = () => {
+  const trackTileStyles = useThemedStyles(createStyles)
+  return (
+    <TrackTileContainer>
+      <View style={styles.mainContent}>
+        <View style={styles.metadata}>
+          <View style={[trackTileStyles.imageContainer, trackTileStyles.image]}>
+            <DynamicImage
```
dylanjeffers, line 38: would it make sense to have an explicit skeleton image component?

src/hooks/useSelectorWeb.ts
```diff
@@ -4,9 +4,11 @@ import { Selector, useSelector } from 'react-redux'
 // When mobile client is no longer dependent on the web client
 // calls to useSelectorWeb can be replaced with useSelector
 export const useSelectorWeb = <ReturnValue>(
-  selector: Selector<CommonState, ReturnValue>
+  selector: Selector<CommonState, ReturnValue>,
+  equalityFn?: (left: ReturnValue, right: ReturnValue) => boolean
```
dylanjeffers, line 8: nice

dylanjeffers (commented):
All looks great so far, going to run locally now

src/components/animated-button/AnimatedButtonProvider.tsx
```diff
@@ -47,7 +47,10 @@ const AnimatedButton = ({
     if (!isPlaying) {
       setIconIndex(i)
     }
-  }, [i, setIconIndex, isPlaying])
+    // Intentionally omitting `isPlaying` from the dependencies,
+    // this effect should only trigger when `i` is updated
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [i, setIconIndex])
```
rickyrombo, line 53: This seems like a potential future foot-gun if not a current bug:
https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528

Maybe use a ref here to track previous `i` values and compare?

src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+// tile height + margin
+const totalTileHeight = {
+  main: 152 + 16,
+  section: 124 + 16,
+  condensed: 124 + 8,
+  playlist: 350
+}
+
+// Helpr to calculate an item count based on the Lineup variant and a multiplier
```
raymondjacobson, line 50: Helpr

src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+            showArtistPick={showLeadingElementArtistPick && !!leadingElementId}
+            showRankIcon={index < rankIconCount}
+            togglePlay={togglePlay}
+            uid={item.uid}
+          />
+        </View>
+      )
+    } else {
+      return null
+      // TODO: Playlist tile
```
raymondjacobson, line 220: 🍝

src/components/lineup/Lineup.tsx
```diff
@@ -0,0 +1,295 @@
...
+    pageItemCount
+  ])
+
+  return (
+    <SectionList
+      ListHeaderComponent={header}
+      ListFooterComponent={<View style={{ height: 160 }} />}
+      onEndReached={onEndReached}
+      onEndReachedThreshold={LOAD_MORE_THRESHOLD}
+      // TODO: Either style the refreshing indicator or
```
raymondjacobson, line 281: This will be fun. Our existing pull to refresh never was perfect on the web view, but I think RN is a great opportunity to make it super slick. Potentially @dylanjeffers could tackle with the profile page pull-to-refresh since that has some nice interplay with the cover photo expanding / fading

src/components/lineup/delineate.ts
```diff
@@ -0,0 +1,56 @@
...
+import { LineupItem } from './types'
+
+const NOW = moment()
+const START_OF_DAY = moment(NOW).startOf('day')
+const START_OF_YESTERDAY = moment(NOW).subtract(1, 'days').startOf('day')
+const START_OF_WEEK = moment(NOW).startOf('week')
+const START_OF_LAST_WEEK = moment(NOW).subtract(1, 'week').startOf('week')
+const START_OF_MONTH = moment(NOW).startOf('month')
+
+const Delineations = Object.freeze({
```
raymondjacobson, line 13: Wonder if we could share this code b/w web & mobile at some point (maybe when we rewrite web lineup)

src/components/lineup/types.ts
```diff
@@ -0,0 +1,108 @@
...
+import Kind from 'audius-client/src/common/models/Kind'
+import { Lineup as LineupData } from 'audius-client/src/common/models/Lineup'
+import { LineupActions } from 'audius-client/src/common/store/lineup/actions'
+import { Maybe } from 'audius-client/src/common/utils/typeUtils'
+import { SectionListProps } from 'react-native'
+
+export enum LineupVariant {
+  MAIN = 'main',
+  SECTION = 'section',
+  CONDENSED = 'condensed',
```
raymondjacobson, line 11: I'm not sure we have all these types on mobile. I think just "main" and "playlist" are the interesting ones

src/components/lineup/types.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Function called to pause playback
+   */
+  pauseTrack: () => void
+
+  /**
+   * Function called to play a track
+   */
+  playTrack: (uid: UID) => void
+
+  /** How many icons to show for top ranked entries in the lineup. Defaults to 0, showing none */
```
raymondjacobson, line 90: nit newline

src/components/track-tile/TrackTile.tsx
```diff
@@ -1,106 +1,215 @@
...
+  GestureResponderEvent,
+  Pressable,
+  StyleSheet
+} from 'react-native'
+import { useDispatch, useSelector } from 'react-redux'
 
+import { BaseStackParamList } from 'app/components/app-navigator/types'
 import { TrackTileProps } from 'app/components/track-tile/types'
+import { useDispatchWeb } from 'app/hooks/useDispatchWeb'
+// import { usePushRouteWeb } from 'app/hooks/usePushRouteWeb'
```
raymondjacobson, line 40: comment

src/components/track-tile/TrackTile.tsx
```diff
@@ -1,106 +1,215 @@
...
-  goToArtistPage,
-  hasCurrentUserReposted,
-  hasCurrentUserSaved,
-  id,
+export const TrackTile = (props: TrackTileProps) => {
+  const { uid } = props
+
+  // Using isEqual as the equality function to prevent rerenders due to object references
+  // not being preserved when syncing redux state from client.
+  // This can be removed when no longer dependent on web client
```
raymondjacobson, line 66: Maybe we should come up with a common AUD-XYZ ticket to capture this work so we know where we left all these todos

src/components/track-tile/TrackTile.tsx
raymondjacobson: This is kind of a "null" guard. @piazzatron built an awesome pattern for this in audius-client, that maybe we should adopt (I ran into similar w/ play bar).

https://github.com/AudiusProject/audius-client/blob/5b0b7683d6fda5148222659d5014356afefb26ec/src/pages/remixes-page/components/desktop/RemixesPage.tsx#L37

The concept is that it captures the nulls so you don't have type issues downstream

src/components/track-tile/TrackTileSkeleton.tsx
raymondjacobson: Or maybe just a <View> with some styling on it versus an image in general?

raymondjacobson (commented):
re: `Question: Do we still need to account for deleted tracks coming back from the api when paginating?`

We shouldn't need to worry about this (yet). The only cases we show deleted tracks are in playlists (which are also lineups, but they aren't paginated). There is a world where in the future we do some smart table reconstruction and paginate the favorites page, etc. and we will need to revisit.

src/components/track-tile/TrackTile.tsx
dylanjeffers: word yeah thats what it seemed like, makes sense, thanks!

raymondjacobson (approved):
Approving with some comments -- I think this is generally super slick and happy to iterate on it after code goes in (if others agree!). I know PRs like this can live in purgatory if we're not careful, so LGTM!!

src/components/dynamic-image/DynamicImage.tsx
```diff
@@ -93,7 +123,7 @@ const DynamicImage = ({
 
   useEffect(() => {
     // Skip animation for subsequent loads where the image hasn't changed
-    if (getPrevImage() !== null && getPrevImage() === image) {
+    if (getPrevImage() !== null && isImageEqual(getPrevImage()!, image)) {
```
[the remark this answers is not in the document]
dylanjeffers: just wondering if we should use `!` just cause it could lead to runtime issue yeah?

dylanjeffers (approved):
Looks great on the device, huge changes great job

src/components/lineup/Lineup.tsx
sliptype: It doesn't, just a habit of putting potentially expensive computations in a memo so that they aren't recomputed on every render. But this one isn't expensive so I can remove the memo

src/components/lineup/Lineup.tsx
sliptype: I should be able to remove this memo too right, because it's just a number and won't trigger rerender unless it changes

src/components/lineup/Lineup.tsx
sliptype: Agreed!

src/components/lineup/delineate.ts
sliptype: SectionList expects an array of sections like so 
```ts
        [
          {
            title: '',
            data: items
          }
        ]
```

The logic here is that we still want the section to exist, we just don't want to show a section delineator. Ideally the `renderSectionHeader` prop on SessionList would be passed an index, but it isn't. So I think this is probably the best approach, but agree it's pretty hacky

src/components/track-tile/TrackTile.tsx
sliptype: Yeah I was thinking through using withNullGuard, but with redux hooks we need a separate component anyway. And in that case, withNullGuard essentially just does a null check, which we could easily do in the component like I'm doing here

Maybe this is a place where we don't use redux hooks and opt for `connect` instead?

src/components/track-tile/TrackTile.tsx
sliptype: I left these in because they are needed for the routing todos, think it should be fine

src/components/track-tile/TrackTile.tsx
sliptype: Love it!

src/components/track-tile/TrackTile.tsx
dylanjeffers: i like what you did, personally its easier to read than connect imo

src/components/track-tile/TrackTile.tsx
sliptype: That makes a lot more sense!

src/components/track-tile/TrackTileSkeleton.tsx
sliptype: I opted for an explicit component that can be used in DynamicImage as well! It's just a LinearGradient view

src/components/animated-button/AnimatedButtonProvider.tsx
sliptype: Ok it's a little more verbose, but this works just as well. What do you think? https://github.com/AudiusProject/audius-mobile-client/pull/249/commits/093be697c056edf8d72a6af71cf9c0c74af87f8e#diff-c9dd9b94e8aadccfa0902fa7e196648983db14a833e54b5c9e786f3c8c32680c

src/components/dynamic-image/DynamicImage.tsx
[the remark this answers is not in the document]
sliptype: Fixed the need for this by assigning the result of getPrevImage to a const before the conditional

src/components/lineup/delineate.ts
sliptype: Yes for sure, I kept it separate for now because the web implementation is pretty different. But we can consolidate in the future

src/components/track-tile/TrackTile.tsx
sliptype: I have a ticket for improving the equality checks: https://linear.app/audius/issue/AUD-1364/reduce-useselectorweb-rerenders

And then the work of moving from useSelectorWeb -> useSelector should be a fairly straighforward find/replace

rickyrombo (approved):

PR #253, opened 2022-01-31
[AUD-1414] Forgot password helper QA Fixes

milindNshah:
### Description
Changes
- Active state on forgot your password text no longer has dark black background
- Updated font weight for drawer text
- Increase tap target 
- Update restore account text


src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
```diff
@@ -42,7 +42,7 @@ const createStyles = (themeColors: ThemeColors) =>
     },
 
     restoreAccount: {
-      fontWeight: '700',
+      fontWeight: '600',
```
sliptype, line 45: I'm pretty sure fontWeight has no effect on Android, would be good to double check

raymondjacobson, line 45: We actually have our own <Text> component now for the font weight stuff, so you should be able to just use that instead of hard-coding here https://github.com/AudiusProject/audius-mobile-client/blob/7cd7b684864032ec4b288cb13f87d70705882e1a/src/components/connect-wallets-drawer/ConnectWalletsDrawer.tsx#L58

sliptype (approved):
Looks good! Just one note

raymondjacobson (approved):
Since there were no changes to the package.json file, can we undo the package-lock changes?

src/components/forgot-password-drawer/ForgotPasswordDrawer.tsx
milindNshah: Converted to use weight on the text property instead and works on both android & ios!

PR #254, opened 2022-02-01
Add makeStyles pattern

dylanjeffers:
### Description

Style pattern that tries to unify and extend our theme use when creating styles

Usage:
```tsx
const useStyles = makeStyles(theme => ({
  headerExample: {
    ...theme.typography.h1,
    color: theme.palette.neutral
  },
})

const ExampleComponent = () => {
  const styles = useStyles()
  
  return (
    <Text style={styles.headerExample}>Example Header</Text>
  )
}
```

Pattern draws inspiration from material-ui!

Future versions can allow props to be passed to useStyles to do more custom styling, example:
```tsx
const useStyles = makeStyles((theme, props) => ({
  exampleStyle: {
     color: props.something ? theme.palette.primary : theme.palette.secondary
   }
})

const ExampleComponent = ({something}) => {
  const styles = useStyles({something})
  
  return (
    <Text style={styles.exampleStyle}>Example Text</Text>
  )
}
```

Currently, `palette` and `typography` are the only things currently in our theme, but we can easily add more, like spacing, zIndex, breakpoints, transitions, etc...

### DIFF
--- a/src/styles/typography.ts
+++ b/src/styles/typography.ts
@@ -14,3 +14,11 @@ export const fontByWeight = {
 export const font = (weight: keyof typeof fontByWeight): TextStyle => ({
   fontFamily: fontByWeight[weight]
 })
+
+export const typography: Record<'h1', TextStyle> = {
+  h1: {
+    fontSize: 18,
+    fontWeight: 'bold',
+    fontFamily: fontByWeight.bold
+  }
+}
--- a/src/styles/typography.ts
+++ b/src/styles/typography.ts
@@ -1,6 +1,16 @@
 import { TextStyle } from 'react-native'
 
-export const fontByWeight = {
+type FontWeight =
+  | 'heavy'
+  | 'bold'
+  | 'demiBold'
+  | 'medium'
+  | 'regular'
+  | 'light'
+  | 'thin'
+  | 'ultraLight'
+
+export const fontByWeight: Record<FontWeight, string> = {
   heavy: 'AvenirNextLTPro-Heavy',
   bold: 'AvenirNextLTPro-Bold',
   demiBold: 'AvenirNextLTPro-DemiBold',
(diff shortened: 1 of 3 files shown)


rickyrombo (approved):

src/styles/typography.ts
```diff
@@ -14,3 +14,11 @@ export const fontByWeight = {
 export const font = (weight: keyof typeof fontByWeight): TextStyle => ({
   fontFamily: fontByWeight[weight]
 })
+
+export const typography: Record<'h1', TextStyle> = {
+  h1: {
+    fontSize: 18,
+    fontWeight: 'bold',
```
sliptype, line 21: We shouldn't need the `fontWeight` because it's already specified by the `fontFamily`

src/styles/typography.ts
```diff
@@ -14,3 +14,11 @@ export const fontByWeight = {
 export const font = (weight: keyof typeof fontByWeight): TextStyle => ({
   fontFamily: fontByWeight[weight]
 })
+
+export const typography: Record<'h1', TextStyle> = {
+  h1: {
```
sliptype, line 19: Currently the Text component also sets the color by default to themeColors.neutral, what do you think would be the best way to achieve that using this pattern? 

It seems correct to not have color as part of the typography definitions, so maybe we just specify the color manually every time?

sliptype (approved):
I like this! The styling stuff is already pretty fragmented so if we can adopt this pattern everywhere that will be nice

src/styles/typography.ts
dylanjeffers: great call

src/styles/typography.ts
dylanjeffers: yeah good callouts... i think maybe it makes sense to set a default text color, ill not include it for this pr, but do some research

PR #256, opened 2022-02-01
Upgrade to new react-jsx transform

dylanjeffers:
### Description

Upgrades babel, metro, eslint, and ts configs to take advantage of React 17+ react-jsx transform, which removes need to explicitly import the React object. There are also plans to upgrade `audius-client` but that requires a bit more work, so staring here!


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "module": "commonjs",
     "lib": ["es6", "dom"],
     "allowJs": true,
-    "jsx": "react-native",
+    "jsx": "react-jsx",
```
raymondjacobson, line 7: I'm surprised "this just works"

raymondjacobson (approved):

PR #258, opened 2022-02-02
Add Tile component, establish core components directory

dylanjeffers:
### Description

- Implements Tile component and refactors TrackTile to use it. It will further be used for other tile components, particularly in the profile screen.
- Establishes a "core" component directory that will contain atomic components used throughout the app.


sliptype (approved):
Looks great!

src/components/track-tile/TrackTileSkeleton.tsx
```diff
@@ -29,27 +26,25 @@ const styles = StyleSheet.create({
...
-          <View style={[trackTileStyles.titles]}>
-            <View style={trackTileStyles.title}>
-              <Skeleton style={styles.skeleton} width='80%' height='80%' />
-            </View>
-            <View style={[trackTileStyles.artist, { width: '100%' }]}>
-              <Skeleton style={styles.skeleton} width='60%' height='80%' />
-            </View>
+        <View style={[trackTileStyles.titles]}>
+          <View style={trackTileStyles.title}>
+            <Skeleton style={styles.skeleton} width='80%' height='80%' />
```
milindNshah, line 37: I think it would improve the 'breakability' of the code if we could pass the width and height as numbers

milindNshah (approved):
looks good!

src/components/track-tile/TrackTileSkeleton.tsx
dylanjeffers: yeah totally agree! i can look into this when i re-use the track tile for profile page. at least for this pr, it looks changed when really it was just a reformat cause the outer view changed

PR #259, opened 2022-02-02
Update DynamicImage styles, rename image to source

dylanjeffers:
### Description
- Re-routes the `style` prop to the root container
- Adds `styles` prop which includes entries for the `root`, `imageContainer`, and `image` sub elements
- Renames `image` prop to `source` to mimic the react-native Image props
- Adds `children` prop, to allow overlays
- Minor perf improvements with `useCallback`
- Moves component to "core" directory

### DIFF
--- a/src/components/core/DynamicImage/DynamicImage.tsx
+++ b/src/components/core/DynamicImage/DynamicImage.tsx
@@ -1,4 +1,11 @@
-import { memo, useCallback, useEffect, useRef, useState } from 'react'
+import {
+  memo,
+  ReactNode,
+  useCallback,
+  useEffect,
+  useRef,
+  useState
+} from 'react'
 
 import transparentPlaceholderImg from 'audius-client/src/common/assets/image/1x1-transparent.png'
 import useInstanceVar from 'audius-client/src/common/hooks/useInstanceVar'
@@ -32,6 +39,8 @@ export type DynamicImageProps = {
   immediate?: boolean
   // Whether or not to use the default placeholder
   usePlaceholder?: boolean
+  // overlays rendered above image
+  children?: ReactNode
 }
 
 const styles = StyleSheet.create({
@@ -101,7 +110,8 @@ const DynamicImage = ({
   style,
   styles: stylesProp,
   immediate,
-  usePlaceholder = true
+  usePlaceholder = true,
+  children
 }: DynamicImageProps) => {
   const [firstSize, setFirstSize] = useState(0)
   const [secondSize, setSecondSize] = useState(0)
@@ -197,6 +207,7 @@ const DynamicImage = ({
           usePlaceholder={usePlaceholder}
         />
       </Animated.View>
+      {children}
     </View>
   )
 }
--- a/src/components/core/DynamicImage/DynamicImage.tsx
+++ b/src/components/core/DynamicImage/DynamicImage.tsx
@@ -10,18 +10,24 @@ import {
   ImageSourcePropType,
   ImageStyle,
   ImageURISource,
+  LayoutChangeEvent,
   StyleProp,
   StyleSheet,
-  View
+  View,
+  ViewStyle
 } from 'react-native'
 
 import { ImageSkeleton } from 'app/components/image-skeleton'
 
 export type DynamicImageProps = {
   // Image source
-  image?: ImageSourcePropType
-  // Style to apply to the image itself
-  style?: StyleProp<ImageStyle>
+  source?: ImageSourcePropType
+  styles?: {
+    root?: ViewStyle
+    imageContainer?: ViewStyle
+    image?: StyleProp<ImageStyle>
+  }
+  style?: ViewStyle
   // Whether or not to immediately animate
   immediate?: boolean
   // Whether or not to use the default placeholder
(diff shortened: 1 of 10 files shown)


src/components/core/DynamicImage/DynamicImage.tsx
```diff
@@ -10,26 +10,32 @@ import {
...
 import { ImageSkeleton } from 'app/components/image-skeleton'
 
 export type DynamicImageProps = {
   // Image source
-  image?: ImageSourcePropType
-  // Style to apply to the image itself
-  style?: StyleProp<ImageStyle>
+  source?: ImageSourcePropType
+  styles?: {
+    root?: ViewStyle
```
Kyle-Shanks, line 26: Can update these to allow for array of styles
```suggestion
    root?: StyleProp<ViewStyle>
```

dylanjeffers: ah yeah great call, is that what StyleProp does basically?

Kyle-Shanks (approved):
Nice 🔥

src/components/core/DynamicImage/DynamicImage.tsx
Kyle-Shanks: Yea that's my understanding. Sebastian mentioned it before and I believe him haha

src/components/core/DynamicImage/DynamicImage.tsx
```diff
@@ -10,26 +17,34 @@ import {
...
+    imageContainer?: ViewStyle
+    image?: StyleProp<ImageStyle>
+  }
+  style?: ViewStyle
```
sliptype, line 37: Should we maybe just get rid of the `style` prop? It might create confusion, what are your thoughts?

src/components/core/DynamicImage/DynamicImage.tsx
```diff
@@ -114,61 +141,77 @@ const DynamicImage = ({
...
       </Animated.View>
+      {children}
     </View>
   )
 }
 
-export default memo(DynamicImage)
+const MemoizedDynamicImage = memo(DynamicImage)
+
+export default MemoizedDynamicImage
```
sliptype, line 217: We could just wrap the component itself in `memo` and export that as a named export

sliptype (approved):
Awesome! All great changes

src/components/core/DynamicImage/DynamicImage.tsx
dylanjeffers: yeah good q, i typically leave it cause its quite common to just want to update the root element, like adding padding, and it feels a bit over-engineered to always use the styles prop. thats just my experience, using className vs classes prop on the web. what do you think?

src/components/core/DynamicImage/DynamicImage.tsx
dylanjeffers: i think what happens there is that we get an anonymous inner function component, ill experiment and paste some results here

src/components/core/DynamicImage/DynamicImage.tsx
sliptype: I agree with that, my only worry is that it's a bit of boiler plate having to define both props on components. Maybe we make a helper type that encapsulates that? Something like: 

```ts
type Style = StyleProp<ViewStyle | ImageStyle | TextStyle>

type PropsWithStyles<
  Props extends Record<string, any>,
  Styles extends { root: Style } & Record<string, Style>
> = Props & {
  style: Styles['root']
  styles: Styles
}

export type DynamicImageProps = PropsWithStyles<
  {
    // Image source
    image?: ImageSourcePropType
    // Whether or not to immediately animate
    immediate?: boolean
    // Whether or not to use the default placeholder
    usePlaceholder?: boolean
  },
  {
    root: StyleProp<ImageStyle>
  }
>
```

src/components/core/DynamicImage/DynamicImage.tsx
dylanjeffers: So wrapping the component itself: 
image
And wrapping afterward:
image

sliptype: Ah ok, makes sense. I do think we landed on wrapping the component with memo in one of our product eng syncs, I see us doing it on a few components including Scrubber. We should probably discuss again today so we avoid those Anonymous comps

dylanjeffers: A way to do it inline is this: `const DynamicImage = memo(function DynamicImage({....}) {...}` what do you think about that?

sliptype: I actually kinda like that, we follow a similar pattern in stems with forwardRefs https://github.com/AudiusProject/stems/blob/master/src/components/Button/Button.tsx#L30

dylanjeffers: okay sweet, can def update with a follow up pr here that adds profile page!

src/components/core/DynamicImage/DynamicImage.tsx
dylanjeffers: ohh wow love this! adding

PR #264, opened 2022-02-04
[AUD-1450] Add profile-page tab-navigation

dylanjeffers:
### Description

Implements tab-navigation for profile-page, based on user's tier level and whether they are an artist or user.

Also updates TopTabNavigation component to allow passing screens in directly and also improves navigation icon flow

### Dragons
Removed the absolute View wrapping the tab navigator and verified that it works okay. but I might be missing something there.


src/components/app-navigator/TopTabNavigator.tsx
```diff
@@ -1,60 +1,77 @@
...
+
 import { createMaterialTopTabNavigator } from '@react-navigation/material-top-tabs'
-import { View } from 'react-native'
 
 import { TopTabBar } from 'app/components/top-tab-bar'
 
 const Tab = createMaterialTopTabNavigator()
 
-type ScreenInfo = {
+const screenOptions = {
```
dylanjeffers, line 9: does it make sense to get the colors and styles from a useStyles?

src/components/top-tab-bar/TopTabBar.tsx
```diff
@@ -143,11 +143,7 @@ export const TopTabBar = ({ state, descriptors, navigation, position }) => {
               style={styles.tab}
               testID={options.tabBarTestID}
             >
-              {Icon && (
-                <Animated.View style={{ opacity }}>
-                  <Icon />
-                </Animated.View>
-              )}
+              <Animated.View style={{ opacity }}>{icon}</Animated.View>
```
Kyle-Shanks, line 146: Could have sworn that I tried this. Nice

src/components/app-navigator/TopTabNavigator.tsx
```diff
@@ -1,60 +1,77 @@
+import { ComponentType, ReactNode } from 'react'
```
Kyle-Shanks, line 1: I like the updates here. Does the component still render and fill the screen as it needs to?

Kyle-Shanks (commented):
Looks good to me pending confirmation that the navigator and screens still render and fill the screen properly

src/components/app-navigator/TopTabNavigator.tsx
sliptype: I would say yes

src/screens/profile-screen/ProfileTabNavigator.tsx
```diff
@@ -27,37 +36,67 @@ const CollectiblesTab = () => {
...
+  profile: ProfileUser
+}
+
+export const ProfileTabNavigator = ({ profile }: ProfileTabNavigatiorProps) => {
+  const { neutral } = useThemeColors()
+  const isArtist = profile.track_count > 0
+
+  const showCollectiblesTab = useShouldShowCollectiblesTab(profile)
+
+  const IconRepostFixed = useCallback(() => <IconRepost fill={neutral} />, [
```
sliptype, line 49: Why is this only needed for the repost icon?

src/screens/profile-screen/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Encapsulates complex logic for determining whether we should show the
+ * profile's collectibles tab.
+ *
+ * User needs to be at a specific tier, needs to have collectibles,
+ * they need to be ordered, and they need to load correctly.
+ *
+ * If the user is viewing their own profile, the collectibles don't
+ * need to be ordered
+ */
+export const useShouldShowCollectiblesTab = (profile: ProfileUser) => {
```
sliptype, line 41: IMO this is the kind of logic that should be shared in common, feels weird to maintain this business logic twice

sliptype (approved):
Good stuff! Couple of comments

src/components/top-tab-bar/TopTabBar.tsx
dylanjeffers: the is it a function, is it a component, is it an element had me tripping for a while :D

src/components/app-navigator/TopTabNavigator.tsx
dylanjeffers: yeah i ran through its usage in profile page and explore page and it seemed to still render in the right way. let's keep tabs on this though, no pun intented!

dylanjeffers:
> Looks good to me pending confirmation that the navigator and screens still render and fill the screen properly

so here is it in profile page: 
image
and in explore: 
image
that's correct yeah?

src/screens/profile-screen/ProfileTabNavigator.tsx
dylanjeffers: for some reason the icon repost svg wasn't picking up the fill, was seeing: 
image
any thoughts there?

src/screens/profile-screen/utils.ts
dylanjeffers: yeah i def agree, mind if combine that logic at a later point?

src/screens/profile-screen/ProfileTabNavigator.tsx
sliptype: It looks like it's because we have the fill defined as `#000` in the repost icon, so we can use the fill prop in other places via react-native-svg. But react-navigation is styling the svgs differently, so the ones that have an actual color defined and aren't being overridden work. 

We should agree on one method of coloring svgs, I'm for react-native-svg (mainly because I don't know how the other method works. Can we just define a fill in the styles?)

src/screens/profile-screen/utils.ts
sliptype: Sounds good!

Kyle-Shanks:
> > Looks good to me pending confirmation that the navigator and screens still render and fill the screen properly
> 
> that's correct yeah?

Yea I was wondering about the tab screen height mostly, but looks good from what I can see here

Kyle-Shanks (approved):

src/screens/profile-screen/ProfileTabNavigator.tsx
dylanjeffers: updated all the icons, and using the provided "color" prop for the icon option, thanks for your help on this

PR #265, opened 2022-02-04
[AUD-1409] Add profile-page tracks and reposts tabs

dylanjeffers:
### Description

Adds tracks and reposts tabs, leveraging the Lineup component

### Dragons

There was a weird bug in `Artwork.tsx` where the dominant color values where not integers and instead something like 120.0000039, which caused a rendering error for the shadow component. Added a workaround using let syntax, curious if folks like that approach vs a purely functional one

### DIFF
--- a/src/screens/profile-screen/EmptyTab.tsx
+++ b/src/screens/profile-screen/EmptyTab.tsx
@@ -13,7 +13,6 @@ const useStyles = makeStyles(({ spacing }) => ({
     paddingHorizontal: spacing(6)
   },
   emptyTabContent: {
-    display: 'flex',
     alignItems: 'center'
   }
 }))
--- a/src/screens/profile-screen/EmptyTab.tsx
+++ b/src/screens/profile-screen/EmptyTab.tsx
@@ -1,0 +1,38 @@
+import { ReactNode } from 'react'
+
+import { Tile } from 'app/components/core'
+import { makeStyles } from 'app/styles'
+
+const useStyles = makeStyles(({ spacing }) => ({
+  emptyTabRoot: {
+    marginVertical: spacing(2),
+    marginHorizontal: spacing(3)
+  },
+  emptyTab: {
+    paddingVertical: spacing(8),
+    paddingHorizontal: spacing(6)
+  },
+  emptyTabContent: {
+    display: 'flex',
+    alignItems: 'center'
+  }
+}))
+
+type EmptyTabProps = {
+  children: ReactNode
+}
+
+export const EmptyTab = ({ children }: EmptyTabProps) => {
+  const styles = useStyles()
+  return (
+    <Tile
+      styles={{
+        root: styles.emptyTabRoot,
+        tile: styles.emptyTab,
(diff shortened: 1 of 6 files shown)


src/screens/profile-screen/RepostsTab.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <EmptyTab>
+        <Text>{"You haven't reposted anything yet"}</Text>
+      </EmptyTab>
+    )
+  }
+
+  return (
+    <Lineup
+      actions={feedActions}
+      lineup={lineup as any}
```
dylanjeffers, line 43: going to work on the typing here, current issue is
```
Property '_loading' is missing in type 'LineupStateTrack<{ id: number; }>' but required in type 'LineupItem'.ts(2322)
[types.ts(14, 3): ]()'_loading' is declared here.
```
any immediate thoughts on what to do here?

sliptype: I think we should remove `_loading` from `LineupItem` and do 

```ts
    if ((item as LoadingLineupItem)._loading) {
```

here: https://github.com/AudiusProject/audius-mobile-client/blob/master/src/components/lineup/Lineup.tsx#L194

src/screens/profile-screen/EmptyTab.tsx
```diff
@@ -0,0 +1,38 @@
...
+  emptyTabRoot: {
+    marginVertical: spacing(2),
+    marginHorizontal: spacing(3)
+  },
+  emptyTab: {
+    paddingVertical: spacing(8),
+    paddingHorizontal: spacing(6)
+  },
+  emptyTabContent: {
+    display: 'flex',
```
sliptype, line 16: Weird flex but ok

(not needed)

src/screens/profile-screen/RepostsTab.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const handlePauseTrack = useCallback(() => {
+    dispatchWeb(feedActions.pause())
+  }, [dispatchWeb])
+
+  if (!profile) return null
+
+  if (profile.repost_count === 0) {
+    return (
+      <EmptyTab>
+        <Text>{"You haven't reposted anything yet"}</Text>
```
sliptype, line 35: Let's use a `messages` object

sliptype (approved):
Looking awesome 🔥

src/screens/profile-screen/RepostsTab.tsx
dylanjeffers: mind double checking the route i took?

sliptype: I think that's fine, but `_isLoading` will never exist on a lineup item. I added that field as a flag for use inside the Lineup logic to display extra skeletons

But I can go back and clean that up when tackling the rest of the Lineup changes

dylanjeffers: ohh gotcha

PR #266, opened 2022-02-04
Update image hooks to use dispatchweb

Kyle-Shanks:
### Description

Did a whoops. Fixing the whoops.
Image hooks for mobile now pass useDispatchWeb so that the fetch actions will actually fetch the art

### How Has This Been Tested?
 The images now render 👍


sliptype (approved):
whoops

PR #267, opened 2022-02-04
[AUD-1421] Track Screen download buttons

sliptype:
### Description

This PR:
* Adds the download buttons to the TrackScreen
Simulator Screen Shot - iPhone 13 - 2022-02-04 at 10 38 49
* Triggers a toast when clicking a disabled download button
* Adds a `useToast` hook which just wraps `useContext`. Thoughts on this? Also should we move away from the toast context on mobile?
* Sets the icon color for common buttons

### Dragons

Dependent on corresponding client pr

### How Has This Been Tested?

Tested on ios simulator


src/screens/track-screen/TrackScreenDownloadButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+    paddingHorizontal: 8
+  },
+  icon: {
+    marginRight: 4
+  },
+  label: {
+    fontSize: 14,
+    padding: 0,
+    textTransform: 'uppercase',
+    ...font('bold'),
```
dylanjeffers, line 53: was trying to add that to makeStyles through the `{fontByWeight}` prop, but maybe we just do `{font}` and then can do `...font.bold` or something?

src/screens/track-screen/TrackScreenDownloadButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+    color: palette.neutralLight4
+  }
+}))
+
+const DownloadButton = ({
+  label,
+  state,
+  onClick = () => {}
+}: DownloadButtonProps) => {
+  const { toast } = useToast()
```
dylanjeffers, line 63: yeah down to keep using the useToast hook, and then we can replace the `useContext` part with `dispatch(toast(...))` like in audius-client eventually?

src/screens/track-screen/TrackScreenDownloadButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const styles = useStyles()
+  const requiresFollow = state === ButtonState.REQUIRES_FOLLOW
+  const isProcessing = state === ButtonState.PROCESSING
+  const isDisabled = state === ButtonState.PROCESSING || requiresFollow
+
+  const onPress = useCallback(() => {
```
dylanjeffers, line 70: so not sure if this is standard, but i typically define callbacks as `handle<Action>`, so if its onPress prop, then `handlePress`. ive noticed this can help with name collisions if a prop is called `onPress` and you need to slightly modify it for the child

src/screens/track-screen/TrackScreenDownloadButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+    if (requiresFollow) {
+      toast({ content: messages.followToDownload })
+    }
+
+    if (isDisabled) {
+      return
+    }
+
+    onClick()
+  }, [isDisabled, onClick, requiresFollow, toast])
```
dylanjeffers, line 80: so are we not going to pass `isDisabled` down to the button here cause we actually do want the onPress to fire?

src/screens/track-screen/TrackScreenDownloadButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+    following,
+    useSelector: useSelectorWeb as typeof useSelector
+  })
+
+  const shouldHide = buttons.length === 0
+  if (shouldHide) {
+    return null
+  }
+
+  return <View style={{ marginBottom: 12 }}>{buttons.map(DownloadButton)}</View>
```
dylanjeffers, line 155: I thinkk we get an issue with key missing here?

src/screens/track-screen/TrackScreenDownloadButtons.tsx
```diff
@@ -0,0 +1,156 @@
...
+        make({
+          eventName: Name.TRACK_PAGE_DOWNLOAD,
+          id,
+          category,
+          parent_track_id: parentTrackId
+        })
+      )
+    },
+    [dispatchWeb, user]
+  )
```
dylanjeffers, line 141: space! adding as a note for eslint

src/screens/track-screen/TrackScreenDownloadButtons.tsx
sliptype: I like that! No reason for it to be a function

dylanjeffers (approved):
nice job! things really coming together

src/screens/track-screen/TrackScreenDownloadButtons.tsx
sliptype: Yup, I need to get in the habit of this. Will update!

src/screens/track-screen/TrackScreenDownloadButtons.tsx
sliptype: Whoops, good catch!

src/screens/track-screen/TrackScreenHeader.tsx
```diff
@@ -47,11 +47,12 @@ import { make, track } from 'app/utils/analytics'
 import { moodMap } from 'app/utils/moods'
 import { ThemeColors } from 'app/utils/theme'
 
+// import HiddenTrackHeader from '../HiddenTrackHeader'
```
milindNshah, line 50: I think this should be deleted

milindNshah (approved):
Looks good!

PR #270, opened 2022-02-07
[AUD-1459] Refactor card lists to FlatList

dylanjeffers:
### Description

Defines `CardList` using `FlatList`, creates `CollectionList` and `ArtistList` using `CardList`, and shares those components in the profile and explore screens.


src/components/collection-list/CollectionList.tsx
```diff
@@ -0,0 +1,37 @@
+import { UserCollection } from 'audius-client/src/common/models/Collection'
+import { Text } from 'react-native'
+
+import { CardList, CardListProps } from 'app/components/core'
+import { CollectionCard } from 'app/screens/explore-screen/components/CollectionCard'
+
+import { EmptyTab } from '../../screens/profile-screen/EmptyTab'
```
sliptype, line 7: Should we move/rename the EmptyTab component?

src/components/core/CardList.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+type CardListRenderItemInfo<ItemT> = ListRenderItemInfo<ItemT> & {
+  style: StyleProp<ViewStyle>
+}
+
+export type CardListProps<ItemT> = ListProps<ItemT> & {
+  renderItem: (info: CardListRenderItemInfo<ItemT>) => ReactElement | null
+}
+
+export function CardList<ItemT>(props: CardListProps<ItemT>) {
```
sliptype, line 22: Did you make this a named function to allow for the type parameter? Another option (not necessarily better) I've discovered is:

```suggestion
export const CardList = <ItemT, _>(props: CardListProps<ItemT>) => {
```

The unused type param lets JSX figure out that it's not a component

src/screens/profile-screen/AlbumsTab.tsx
```diff
@@ -1,6 +1,9 @@
+import { useMemo } from 'react'
+
 import { useSelectorWeb } from 'app/hooks/useSelectorWeb'
 
-import { CollectionList } from './CollectionList'
+import { CollectionList } from '../../components/collection-list/CollectionList'
```
sliptype, line 5:
```suggestion
import { CollectionList } from 'app/components/collection-list/CollectionList'
```

src/screens/profile-screen/AlbumsTab.tsx
```diff
@@ -9,12 +12,17 @@ const messages = {
 
 export const AlbumsTab = () => {
   const { profile, albums } = useSelectorWeb(getProfile)
-  if (!profile || !albums) return null
+
+  const userAlbums = useMemo(() => {
+    if (profile && albums)
```
sliptype, line 17: I prefer curlies for single line conditionals, but I don't think we have that in the styleguide

src/styles/typography.ts
```diff
@@ -41,5 +46,10 @@ export const typography = {
     fontFamily: fontByWeight.medium,
     marginBottom: 14 * 0.35
   },
+  body2: {
```
sliptype, line 49: `body2` seems kinda ambiguous to me. I guess it does make sense following h1, h2, etc. pattern though

sliptype (approved):
Love this consolidation

src/components/collection-list/CollectionList.tsx
dylanjeffers: yeah def! will rename to EmptyListItem i guess?

src/components/core/CardList.tsx
dylanjeffers: totally, its just like... ugh why ts... :/ can update if that's the standard!

src/screens/profile-screen/AlbumsTab.tsx
dylanjeffers: yeah agreed, was missed!

src/styles/typography.ts
dylanjeffers: yeah its... not the best, im kinda throwing out values here to try and categorize stuff, just so we can take stock of which values we are centering on, and then we can bring to julian to nail down naming convention?

src/components/core/CardList.tsx
sliptype: It's not a standard! both ways are annoying for sure

PR #271, opened 2022-02-07
Refactor Card to use Tile, update Tile shadows

dylanjeffers:
### Description

- Refactors Card component to use Tile
- Simplifies CardList impl since we wrap a View now instead of passing style through to card. This is needed cause padding directly on the card Tile messes with shadow.
- Updates tile shadow to more closely align with web client tile shadows


sliptype (approved):
lgtm!

PR #274, opened 2022-02-08
[AUD-1464] Merge native and mobile navigation

dylanjeffers:
### Description

First pass at merging native and web navigation into a single hook. Maintains strong typing for native navigation, and we can work on a similar strong typing for web too.

### DIFF
--- a/src/components/track-tile/TrackTile.tsx
+++ b/src/components/track-tile/TrackTile.tsx
@@ -1,7 +1,5 @@
 import { useState, useEffect, useRef, useCallback } from 'react'
 
-import { useNavigation } from '@react-navigation/native'
-import { NativeStackNavigationProp } from '@react-navigation/native-stack'
 import { Track } from 'audius-client/src/common/models/Track'
 import { User } from 'audius-client/src/common/models/User'
 import { getUserId } from 'audius-client/src/common/store/account/selectors'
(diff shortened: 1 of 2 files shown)


src/hooks/useNavigation.ts
```diff
@@ -0,0 +1,43 @@
...
+> = {
+  native: { screen: RouteName; params: ParamList[RouteName] }
+  web: {
+    route: string
+    fromPage?: string
+    fromNativeNotifications?: string
+  }
+}
+
+type NavigationType = NativeStackNavigationProp<BaseStackParamList>
```
sliptype, line 23: `BaseStackParamList` might not be extensive enough for all our stacks, so we might need a type parameter for that at some point. But probably fine for now

src/components/track-tile/TrackTile.tsx
```diff
@@ -107,14 +98,19 @@ const TrackTileComponent = ({
...
-    [pushRouteWeb, permalink, navigation, track_id]
+    [navigation, permalink, track_id]
   )
 
   const goToArtistPage = (e: GestureResponderEvent) => {
-    navigation.navigate('profile', { handle: user.handle })
+    navigation.navigate({
+      native: { screen: 'profile', params: { handle: user.handle } },
+      web: { route: user.handle }
+    })
```
sliptype, line 113: I like this! My only thought is that there will be a fair amount of changes when we rip out web. But that should be fine

sliptype (approved):
Looks great! Happy that the long navigation type gets hidden away too

src/components/track-tile/TrackTile.tsx
dylanjeffers: yeah that is a really good call...

src/hooks/useNavigation.ts
dylanjeffers: word, can def try and extend as we use it in more complex places!

PR #278, opened 2022-02-09
[AUD-971] Add App TopBar

dylanjeffers:
### Description

Adds a basic header that can open the notification panel, route to trending page, and open search page. Not a final solution, since where we put the header in our navigation stack is still up for debate i think.
image


src/components/app-navigator/Header.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const navigation = useNavigation()
+
+  const handlePressHome = useCallback(() => {
+    navigation.navigate({
+      native: { screen: 'trending', params: undefined },
+      web: { route: 'trending' }
+    })
+  }, [navigation])
+
+  if (children === 'profile') {
```
rickyrombo, line 50: I'm not super familiar with `Stack.Navigator`, how does `children` get passed through here?

rickyrombo (approved):

src/components/app-navigator/Header.tsx
dylanjeffers: yeah it's quite weird, so basically the "title" of each screen, which in our case is the "name" prop of each screen, gets passed as the "children" prop in the "headerTitle" prop

rickyrombo: ahh gotcha, I was looking at the docs for `header` not `headerTitle`. Any thoughts on using one over the other?

dylanjeffers: would be down for `header` where is that in the docs?

rickyrombo: was reading stuff here: https://reactnavigation.org/docs/stack-navigator#header-related-options

dylanjeffers: ohh this is exactly what we need, much better that the other prop, great find.

Kyle-Shanks (approved):
🔥

sliptype (approved):
Nice! 

Currently we use a context in mobile web to determine which items to show in the header, I was thinking we could move that logic into redux so it can be shared in common. Future scope, just as an fyi

src/components/app-navigator/Header.tsx
```diff
@@ -0,0 +1,92 @@
+import { useCallback } from 'react'
+
```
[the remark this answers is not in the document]
dylanjeffers: yeah these are great points, i was just going with the react-navigation language here, but topbar makes sense for consistency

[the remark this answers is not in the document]
dylanjeffers: @sliptype i am also curious what the best move here is regarding context vs redux. it seems that different stacks can render different headers, so maybe the move is to let the screens define them? it might be a bit more direct and programatic that way. def need to look into what way is most standard and efficient tho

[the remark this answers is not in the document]
sliptype: That's a great idea, the current implementation isn't very declarative

PR #280, opened 2022-02-09
[AUD-1466] Add follow/following profile buttons

dylanjeffers:
### Description

Adds follow/following buttons
Adds a button component that correctly maps to "primary" and "secondary" buttons like we see in stems. not sure if "secondary" is actually different for mobile


sliptype (approved):
We should just update the main button component's secondary variant to match stems. The place where I'm currently using it is very much a one-off, I can go back and fix it

PR #286, opened 2022-02-11
Add TextButton

dylanjeffers:
### Description

- Adds `TextButton` component for buttons that are text only, which we use in a few places. Discussed this component type with Julian, and this pr is an initial implementation, handling primary and secondary styles.
- Also removes marginBottom (gutters) for body typography


src/components/core/TextButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+      ...typography.body
+    }
+  }
+
+  return merge(baseStyles, variantStyles[variant])
+})
+
+type TextButtonProps = PressableProps &
+  ButtonProps & {
+    variant: 'primary' | 'secondary'
```
rickyrombo, line 31: We use `type` for the button component, should they be consistent? If so, I'd vote for `variant` everywhere because type is so overloaded (and my webdev brain thinks of `<input type="submit">` etc)

src/components/core/TextButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+export const TextButton = (props: TextButtonProps) => {
+  const { title, variant, ...other } = props
```
rickyrombo, line 35: more of a comment on ReactNative, but why did they choose to call the text content `title`? It breaks my web dev brain :(

rickyrombo (approved):

src/components/core/TextButton.tsx
dylanjeffers: yes i also agree with variant, since type is overloaded, that was the intention here! adding a linear for moving everything to variant

src/components/core/TextButton.tsx
dylanjeffers: no clue, i would have used text, or just children itself for buttons, but react-native decided, and id rather keep the prop convention the same.....

rickyrombo: should keep it the same yes 💯  just complaining haha

dylanjeffers: totally

PR #287, opened 2022-02-11
Add headerLeft and headerRight to TopBar

dylanjeffers:
### Description

Adds support for `headerRight` and headerLeft`, options for react-navigation screens. This allows screens to define their own headerLeft and headerRight, using config, or using the navigation context.

### DIFF
--- a/src/components/app-navigator/TopBar.tsx
+++ b/src/components/app-navigator/TopBar.tsx
@@ -38,6 +38,18 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
     paddingRight: 20,
     marginBottom: spacing(1)
   },
+  headerLeft: {
+    justifyContent: 'center',
+    alignItems: 'flex-start',
+    height: spacing(6),
+    width: spacing(20)
+  },
+  headerRight: {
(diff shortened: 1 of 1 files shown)


src/components/app-navigator/TopBar.tsx
```diff
@@ -82,28 +97,38 @@ export const TopBar = memo(
...
-            <IconButton
-              icon={IconNotification}
-              styles={{ icon: styles.iconNotification }}
-              fill={neutralLight4}
-              onPress={handlePressNotification}
-            />
-          )}
+          <View style={styles.headerLeft}>
+            {headerLeft ? (
+              headerLeft({})
```
rickyrombo, line 102: What is the type of the args here out of curiousity?

rickyrombo (approved):

src/components/app-navigator/TopBar.tsx
dylanjeffers: oh yeah good call, its like "touchTint", "color", basically things to help the styles of the button when it's clicked. thinking for now we can ignore it, since these items should be buttons anyway that can handle themselves

PR #289, opened 2022-02-12
[AUD-1412] Add EditProfileScreen

dylanjeffers:
### Description

- Adds EditProfileScreen component
- Introduces Formik for form management
- Uses headerRight and headerLeft TopNav options to set `cancel` and `save` buttons


package.json
```diff
@@ -58,6 +58,7 @@
     "audius-client": "github:AudiusProject/audius-client#e3e8c560ddf5d15416ac6bb498c53f6691909b5d",
     "bn.js": "5.2.0",
     "expo-crypto": "9.2.0",
+    "formik": "^2.2.9",
```
sliptype, line 61: 📌

Although I've been thinking, since we are using `npm ci` now we might not need to pin in the package.json. Just would need to be careful about when to commit package-lock

src/components/app-navigator/BottomTabNavigator.tsx
```diff
@@ -66,6 +67,7 @@ const createStackScreen = <StackParamList extends ParamListBase>(
       {baseScreen(Stack)}
       <Stack.Screen name='track' component={TrackScreen} />
       <Stack.Screen name='profile' component={ProfileScreen} />
+      <Stack.Screen name='EditProfile' component={EditProfileScreen} />
```
sliptype, line 70: Should we be consistent on the casing for the screen names?

src/screens/edit-profile-screen/EditProfileScreen.tsx
```diff
@@ -0,0 +1,158 @@
...
+  save: 'Save',
+  cancel: 'Cancel'
+}
+
+const screenHeight = Dimensions.get('window').height
+
+const useStyles = makeStyles(({ palette }) => ({
+  screen: {
+    display: 'flex',
+    flexDirection: 'column',
```
sliptype, line 39: You shouldn't need these two styles

src/screens/edit-profile-screen/launchSelectImageActionSheet.ts
```diff
@@ -0,0 +1,101 @@
...
+      handlePhoto
+    )
+  }
+
+  if (Platform.OS === 'ios') {
+    // iOS ActionSheet
+    ActionSheetIOS.showActionSheetWithOptions(
+      {
+        options: ['Cancel', 'Photo Library', 'Take Photo'],
+        tintColor: '#7E1BCC',
```
sliptype, line 63: Could this color come from the theme?

sliptype (approved):
Looks awesome! We're getting to a level of code reuse that's 👌👌👌

src/screens/edit-profile-screen/EditProfileScreen.tsx
dylanjeffers: yup! thanks

src/screens/edit-profile-screen/launchSelectImageActionSheet.ts
dylanjeffers: yeah def, suppose itll be a lil hard here since this is a function though, what do you think?

src/components/app-navigator/BottomTabNavigator.tsx
dylanjeffers: yes totally agree, was adding this to get your thoughts cause it seems react-navigation prefers pascal case, thoughts? fine with just doing kebab like we have been tho

src/screens/edit-profile-screen/launchSelectImageActionSheet.ts
sliptype: Adding a `color` argument to this function makes sense to me!

src/components/app-navigator/BottomTabNavigator.tsx
sliptype: I actually don't know why I deviated from the react-navigation convention. I'm fine with pascal case!

dylanjeffers:
moving to audius-client monorepo

